PR #73, opened 2015-02-10
Tweak logging output when processing sqs ban messages.

rfk:
This will give slightly more helpful log messages when we fail to process from the sqs ban queue.  @chilts r?


rfk:
I need this on latest but it keeps getting overwritten by the updater; r=me

PR #71, opened 2014-10-14
Add .shouldBlock() and .isRateLimited(); normalise .isBlocked()

chilts:

PR #74, opened 2015-02-13
More code cleanups

rfk:
This addresses a couple of nits we had left over from #71.  @chilts r?

### DIFF
--- a/email_record.js
+++ b/email_record.js
@@ -115,26 +115,27 @@ module.exports = function (RATE_LIMIT_INTERVAL_MS, BLOCK_INTERVAL_MS, MAX_EMAILS
   }
 
   EmailRecord.prototype.update = function (action) {
-    // if this user is not yet blocked
-    // and if this is NOT an email action, then no block
-    if ( !this.isBlocked() && !isEmailSendingAction(action) ) {
-      return 0
-    }
-
-    // check if this is already blocked, don't count any more hits and tell them to retry
-    if (this.shouldBlock()) {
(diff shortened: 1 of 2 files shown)


email_record.js
```diff
@@ -115,26 +115,27 @@ module.exports = function (RATE_LIMIT_INTERVAL_MS, BLOCK_INTERVAL_MS, MAX_EMAILS
...
-
-    if (this.isOverEmailLimit()) {
-      // rate limit this email if now over the limit and tell them to retry
-      this.rateLimit()
-      return this.retryAfter()
+    // For email-sending actions, we may need to rate-limit.
+    if (isEmailSendingAction(action)) {
+      // If they're already being blocked then don't count any more hits,
+      // and tell them to retry.
+      if (this.shouldBlock()) {
```
shane-tomlinson: What is the difference between `shouldBlock` and `isBlocked`?

rfk: It's...complicated.  "should block" == "is explicited marked as blocked" || "has exceeded sending rate limit".  We could probably name these better.

rfk:
Bump for @chilts to take a look at this please, let's make some incremental cleanups in this repo over time.

email_record.js
chilts: +10101 for renaming these things as we see fit. And also, get rid of those pesky really short prefixes in the objects, but that's a rant for another day. :)

chilts:
This all looks fine, mostly mechanical (1st and 3rd commit). Second commit looks cool too. r+

PR #76, opened 2015-02-25
feat(config): Add a badLoginLockoutIntervalSeconds configuration option.

### LINKED ISSUE #75: badLoginLockout should persist for longer than other rate-limiting flags
I'm migrating this from https://github.com/mozilla/fxa-auth-server/issues/801#issuecomment-74300269 where @ckarlof says:

"""
The intention of account locking was to put up a bigger speed bump for attackers who weren't deterred by our gentler rate limiting that automatically resets after rateLimitIntervalSeconds. Meaning, the lockout was meant to address attacks where there are badLoginLockout bad logins over a longer period, e.g. n*rateLimitIntervalSeconds. I don't know what n should be. Maybe 10, or 100? We can tweak that empirically, I suppose.
"""

This would specific help guard against the following style of attack: https://github.com/blog/1698-weak-passwords-brute-forced

The simplest approach is probably just to add a new config option `badLoginLockoutIntervalSeconds` or similar that we can tweak independently of the other options.

shane-tomlinson:
This is the time over which bad logins should be remembered for account lockout

fixes #75

### DIFF
--- a/config/config.js
+++ b/config/config.js
@@ -63,10 +63,16 @@ module.exports = function (fs, path, url, convict) {
         env: 'MAX_BAD_LOGINS'
       },
       badLoginLockout: {
-        doc: 'Number failed login attempts within rateLimitIntervalSeconds before forcing locking the account',
+        doc: 'Number failed login attempts within badLoginLockoutIntervalSeconds before locking the account',
         default: 20,
         format: 'nat',
         env: 'BAD_LOGIN_LOCKOUT'
+      },
+      badLoginLockoutIntervalSeconds: {
+        doc: 'Duration which a failed login attempt should be remembered towards account lockout',
+        default: 60 * 60 * 30, // three hours for now
+        format: 'nat',
+        env: 'BAD_LOGIN_LOCKOUT_INTERVAL_SECONDS'
       }
     },
     memcache: {
--- a/email_record.js
+++ b/email_record.js
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 // Keep track of events tied to just email addresses
-module.exports = function (RATE_LIMIT_INTERVAL_MS, BLOCK_INTERVAL_MS, MAX_EMAILS, BAD_LOGIN_LOCKOUT, now) {
+module.exports = function (RATE_LIMIT_INTERVAL_MS, BLOCK_INTERVAL_MS, BAD_LOGIN_LOCKOUT_INTERVAL_MS, MAX_EMAILS, BAD_LOGIN_LOCKOUT, now) {
 
   now = now || Date.now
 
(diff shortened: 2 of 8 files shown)


email_record.js
```diff
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 // Keep track of events tied to just email addresses
-module.exports = function (RATE_LIMIT_INTERVAL_MS, BLOCK_INTERVAL_MS, MAX_EMAILS, BAD_LOGIN_LOCKOUT, now) {
+module.exports = function (RATE_LIMIT_INTERVAL_MS, BLOCK_INTERVAL_MS, MAX_EMAILS, BAD_LOGIN_LOCKOUT, BAD_LOGIN_LOCKOUT_INTERVAL_MS, now) {
```
shane-tomlinson: Part of me feels like *_INTERVAL_MS should all be grouped together.

config/config.js
```diff
@@ -63,10 +63,16 @@ module.exports = function (fs, path, url, convict) {
...
       badLoginLockout: {
-        doc: 'Number failed login attempts within rateLimitIntervalSeconds before forcing locking the account',
+        doc: 'Number failed login attempts within badLoginLockoutIntervalSeconds before locking the account',
         default: 20,
         format: 'nat',
         env: 'BAD_LOGIN_LOCKOUT'
+      },
+      badLoginLockoutIntervalSeconds: {
+        doc: 'Duration which a failed login attempt should be remembered towards account lockout',
+        default: 60 * 60 * 30, // three hours for now
```
shane-tomlinson: What is a reasonable default?

shane-tomlinson:
@rfk, @chilts and @dannycoates - I want to have another pass at this tomorrow, any feedback is appreciated! My concerns are already highlighted.

test/remote/too_many_bad_logins.js
```diff
@@ -14,7 +14,8 @@ var config = {
     port: 7000
   },
   limits: {
-    rateLimitIntervalSeconds: 1
+    rateLimitIntervalSeconds: 1,
+    badLoginLockoutIntervalSeconds: 1
```
shane-tomlinson: Strictly speaking, this line isn't used for anything as far as I can tell. I added it for consistency.

email_record.js
rfk: :+1:

We should feel free to cleanup/refactor aggressively in this repo as we go.

config/config.js
rfk: No idea, I think the best we can do here is monitor and tweak.  Based on what you have here, we'll trigger after 20 bad logins over three hours == an average of one bad login every nine minutes, which seems a reasonable place to start.

shane-tomlinson:
Removing the WIP label, updated the order of the input params to emailRecord put the `BAD_LOGIN_LOCKOUT_INTERVAL_MS` next to the other intervals.

Calling on @rfk, @dannycoates and @chilts - ready for review!

rfk:
@shane-tomlinson this looks good.  Is it all covered by existing tests so we don't need to add any with this PR?  If so, r+.  We'll just have to remember to cut and ship a new version of this thing with train-33.

config/config.js
jrgm: > default: 60 * 60 * 30, // three hours for now

That's three hours if the unit is deciseconds ;-)

PR #79, opened 2015-04-01
Add task for tagging release and fix some jshint whining

jrgm:
The first commit for test is to quiet a jshint warning for "'obj' is defined but never used". Previously jshint was passing since it was not linting these files. 

The second commit is just like fxa-auth-server and fxa-auth-db-mysql repos, but review the file matching patterns. r? @chilts @pdehaan

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,12 +28,14 @@
   "devDependencies": {
     "ass": "0.0.4",
     "grunt": "0.4.5",
+    "grunt-bump": "0.3.0",
     "grunt-cli": "0.1.13",
     "grunt-contrib-jshint": "0.10.0",
     "grunt-copyright": "0.1.0",
     "grunt-nsp-shrinkwrap": "0.0.3",
     "jshint": "2.5.x",
     "jshint-stylish": "0.4.0",
+    "load-grunt-tasks": "3.1.0",
     "request": "2.40.0",
     "tap": "0.4.12",
     "walk": "2.3.x"
--- a/test/remote/password_reset_tests.js
+++ b/test/remote/password_reset_tests.js
@@ -36,6 +36,7 @@ test(
       function (err, req, res, obj) {
         t.notOk(err, 'good request is successful')
         t.equal(res.statusCode, 200, 'good request returns a 200')
+        t.ok(obj, 'got an obj, make jshint happy')
         t.end()
       }
     )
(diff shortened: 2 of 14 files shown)


jrgm:
I also added an update to shrinkwrap.

package.json
```diff
@@ -28,12 +28,14 @@
   "devDependencies": {
     "ass": "0.0.4",
```
pdehaan: Unrelated: Why does this use **ass@0.0.4** but [mozilla/fxa-auth-db-server](https://github.com/mozilla/fxa-auth-db-server/blob/master/package.json#L24) uses **ass@git://github.com/jrgm/ass.git#5be99ee**?

jrgm: I need to get all of these in line. But to answer, after 0.0.4 some fixes came in, but also added a large set of binary dependencies. jrgm/ass.git#5be99ee undoes that required binary dependency, and keeps the other fixes. So, the coverage numbers are correct in this repo but html display is fugly, since the filenames have full paths and take up half the page.

test/remote/password_reset_tests.js
```diff
@@ -36,6 +36,7 @@ test(
       function (err, req, res, obj) {
         t.notOk(err, 'good request is successful')
         t.equal(res.statusCode, 200, 'good request returns a 200')
+        t.ok(obj, 'got an obj, make jshint happy')
```
pdehaan: After about the 40th time of reading 'make jshint happy', I'm starting to lean towards "relax the option, this is silly".

Maybe change .jshintrc's `unused` to ["vars"](https://jslinterrors.com/unused-a) (per [mozilla/fxa-auth-server](https://github.com/mozilla/fxa-auth-server/blob/master/.jshintrc#L8)). Or counter proposal of creating a test/.jshintrc with that rule disabled.

> `vars` &mdash; Only warn about unused variable and function declarations. Don't warn about any function parameters.
> &mdash; via https://jslinterrors.com/unused-a

---

> This can be set to `vars` to only check for variables, not function parameters, or `strict` to check all variables and parameters. The default (`true`) behavior is to allow unused parameters that are followed by a used parameter.
> &mdash; via http://jshint.com/docs/options/#unused

package.json
pdehaan: :+1: for @jrgm's ass.

pdehaan:
Nits and sexual harassment aside. A :+1:/:-1: from me.
I vote to either relax the "unused" .jshint rule globally in .jshintrc, or override and disable in test/.jshintrc. Having that `t.ok()` repeated 16 times seems non-optimal. 

I'll defer to @chilts and @dannycoates to give the final call.

jrgm:
Yeah, I considered changing the jshint rule. I hate having stuff that only exists to satisfy a linter. For this particular rule, I could go with not having it worry about unused trailing arguments.

PR #86, opened 2015-06-10
chore(shrinkwrap): update ass to what other modules use; update shrinkwrap

jrgm:
I went to update shrinkwrap, and the version of `ass` we use causes a weird error with shrinkwrap. So use the version we use in other repos. 

@chilts r?


jrgm:
Actually, I need to merge this now, and it's just a boilerplate change.

PR #98, opened 2016-02-17
refactor(lib): Put all the code inside a "lib" subdirectory.

rfk:
I was digging into the code here, and realized that it doesn't follow our convention of putting the app code inside a "lib" sub-directory.  This PR moves the code out of the top-level dir and into "lib", and creates a stub file in ./bin/ that imports and runs it.  Functionality it not modified.  @vbudhram r?


vbudhram:
@rfk looks good to me, lets merge this one in before #99

PR #99, opened 2016-02-17
feat(api): Add account status check with rate limiting

### LINKED ISSUE #96: Add support for rate-limiting /account/status endpoint
ref https://github.com/mozilla/fxa-auth-server/issues/1176

vbudhram:
This PR adds a new route to the customs server, `POST /accountStatus`. This is used to rate-limit requests by ip address, for getting an account's status. This is a dependency for https://github.com/mozilla/fxa-auth-server/pull/1184

Fixes #96 

@rfk this is what you had in mind? r?

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,7 @@ There are three types of policies:
 We currently have the following policies in place:
 
 * rate-limiting when too many emails (`config.limits.maxEmails` defaults to 3) have been sent to the same email address in a given time period (`config.limits.rateLimitIntervalSeconds` defaults to 15 minutes)
+* rate-limiting when too many requests to look up account status by email address (`config.limits.maxAccountStatusCheck`) have been sent from the same ip address during period (`config.limits.rateLimitIntervalSeconds` defaults to 15 minutes)
 * rate-limiting when too many failed login attempts (`config.limits.maxBadLogins` defaults to 2) have occurred for a given account and IP address, in a given time period (`config.limits.rateLimitIntervalSeconds` defaults to 15 minutes)
 * lockout when too many failed login attempts (`config.limits.badLoginLockout` defaults to 20) have occurred for a given account regardless of the IP address, in a given time period  (`config.limits.rateLimitIntervalSeconds` defaults to 15 minutes)
 * manual blocking of an account (see `/blockEmail` API call)
--- a/lib/ip_record.js
+++ b/lib/ip_record.js
@@ -3,36 +3,89 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 // Keep track of events related to just IP addresses
-module.exports = function (BLOCK_INTERVAL_MS, now) {
+module.exports = function (BLOCK_INTERVAL_MS, RATE_LIMIT_INTERVAL_MS, MAX_ACCOUNT_STATUS_CHECK, now) {
 
   now = now || Date.now
 
+  var ACCOUNT_STATUS_ACTION = {
+    accountStatusCheck       : true
+  }
+
   function IpRecord() {}
+
+  function isAccountStatusAction(action) {
+    return ACCOUNT_STATUS_ACTION[action]
+  }
 
   IpRecord.parse = function (object) {
     var rec = new IpRecord()
     object = object || {}
     rec.bk = object.bk // timestamp when the account was blocked
+    rec.as = object.as || []  // timestamps when account status check occurred
+    rec.rl = object.rl  // timestamp when the account was rate-limited
     return rec
   }
 
+  IpRecord.prototype.isOverAccountStatusCheck = function () {
+    return this.as.length > MAX_ACCOUNT_STATUS_CHECK
+  }
+
+  IpRecord.prototype.trimAccountStatus = function (now) {
+    if (this.as.length === 0) { return }
+    // "as" is naturally ordered from oldest to newest
+    // and we only need to keep up to MAX_ACCOUNT_STATUS_CHECK + 1
+
+    var i = this.as.length - 1
+    var n = 0
+    var hit = this.as[i]
+    while (hit > (now - RATE_LIMIT_INTERVAL_MS) && n <= MAX_ACCOUNT_STATUS_CHECK) {
+      hit = this.as[--i]
+      n++
+    }
+    this.as = this.as.slice(i + 1)
+  }
+
+  IpRecord.prototype.addAccountStatusCheck = function () {
+    this.trimAccountStatus(now())
+    this.as.push(now())
+  }
+
   IpRecord.prototype.shouldBlock = function () {
-    return this.isBlocked()
+    return this.isBlocked() || this.isRateLimited()
   }
 
   IpRecord.prototype.isBlocked = function () {
     return !!(this.bk && (now() - this.bk < BLOCK_INTERVAL_MS))
   }
 
+  IpRecord.prototype.isRateLimited = function () {
+    return !!(this.rl && (now() - this.rl < RATE_LIMIT_INTERVAL_MS))
+  }
+
   IpRecord.prototype.block = function () {
     this.bk = now()
   }
 
-  IpRecord.prototype.retryAfter = function () {
-    return Math.max(0, Math.floor(((this.bk || 0) + BLOCK_INTERVAL_MS - now()) / 1000))
+  IpRecord.prototype.rateLimit = function () {
+    this.rl = now()
+    this.as = []
   }
 
-  IpRecord.prototype.update = function () {
+  IpRecord.prototype.retryAfter = function () {
+    var rateLimitAfter = Math.floor(((this.rl || 0) + RATE_LIMIT_INTERVAL_MS - now()) / 1000)
+    var banAfter = Math.floor(((this.bk || 0) + BLOCK_INTERVAL_MS - now()) / 1000)
+    return Math.max(0, rateLimitAfter, banAfter)
+  }
+
+  IpRecord.prototype.update = function (action) {
+    // Increment account status check and throttle if needed
+    if (isAccountStatusAction(action)) {
+      this.addAccountStatusCheck()
+      if (this.isOverAccountStatusCheck() && !this.isRateLimited()){
+        this.rateLimit()
+      }
+    }
+
     return this.retryAfter()
   }
 
--- a/test/local/ip_record_tests.js
+++ b/test/local/ip_record_tests.js
@@ -10,7 +10,7 @@ function now() {
 }
 
 function simpleIpRecord() {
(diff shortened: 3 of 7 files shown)


README.md
```diff
@@ -64,6 +64,7 @@ There are three types of policies:
 We currently have the following policies in place:
 
 * rate-limiting when too many emails (`config.limits.maxEmails` defaults to 3) have been sent to the same email address in a given time period (`config.limits.rateLimitIntervalSeconds` defaults to 15 minutes)
+* rate-limiting when too many request for checking account status (`config.limits.maxAccountStatusCheck`) have been sent from the same ip address during period (`config.limits.rateLimitIntervalSeconds` defaults to 15 minutes)
```
rfk: I feel like this could be slightly clearer about what we're guarding against, e.g. "too many requests to look up account status by email address".  We won't rate-limit attempts to look up account status by e.g. uid or sessionToken.

ip_record.js
```diff
@@ -3,36 +3,89 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 // Keep track of events related to just IP addresses
-module.exports = function (BLOCK_INTERVAL_MS, now) {
+module.exports = function (BLOCK_INTERVAL_MS, now, RATE_LIMIT_INTERVAL_MS, MAX_ACCOUNT_STATUS_CHECK) {
```
rfk: I believe `now` is intended to be an optional parameter, specified only when calling these from tests.  In which case we should leave it at the end of the parameter list and insert the new params before it.

bin/customs_server.js
```diff
@@ -16,8 +16,8 @@ var RATE_LIMIT_INTERVAL_MS = config.limits.rateLimitIntervalSeconds * 1000
 var BAD_LOGIN_LOCKOUT_INTERVAL_MS = config.limits.badLoginLockoutIntervalSeconds * 1000
 
 var IpEmailRecord = require('../ip_email_record')(RATE_LIMIT_INTERVAL_MS, config.limits.maxBadLogins)
-var EmailRecord = require('../email_record')(RATE_LIMIT_INTERVAL_MS, BLOCK_INTERVAL_MS, BAD_LOGIN_LOCKOUT_INTERVAL_MS, config.limits.maxEmails, config.limits.badLoginLockout)
-var IpRecord = require('../ip_record')(BLOCK_INTERVAL_MS)
+var EmailRecord = require('../email_record')(RATE_LIMIT_INTERVAL_MS, BLOCK_INTERVAL_MS, BAD_LOGIN_LOCKOUT_INTERVAL_MS, config.limits.maxEmails, config.limits.badLoginLockout, Date.now, config.limits.maxAccountStatusCheck)
```
rfk: AFAICT you're not changing `email_record.js`, you shouldn't need to pass in the extra config param here.

bin/customs_server.js
```diff
@@ -97,7 +97,7 @@ api.post(
 
     if (!email || !ip || !action) {
       var err = {code: 'MissingParameters', message: 'email, ip and action are all required'}
-      log.error({ op: 'request.failedLoginAttempt', email: email, ip: ip, action: action, err: err })
+      log.error({ op: 'request.check', email: email, ip: ip, action: action, err: err })
```
rfk: nice catch

ip_record.js
```diff
@@ -3,36 +3,89 @@
...
     return rec
   }
 
+  IpRecord.prototype.isOverAccountStatusCheck = function () {
+    return this.as.length > MAX_ACCOUNT_STATUS_CHECK
+  }
+
+  IpRecord.prototype.trimAccountStatus = function (now) {
+    if (this.as.length === 0) { return }
+    // as is naturally ordered from oldest to newest
```
rfk: nit: quote `'as'` here so it more obviously reads as a name, not a grammatical error :-)

ip_record.js
```diff
@@ -3,36 +3,89 @@
...
+    var banAfter = Math.floor(((this.bk || 0) + BLOCK_INTERVAL_MS - now()) / 1000)
+    return Math.max(0, rateLimitAfter, banAfter)
   }
 
-  IpRecord.prototype.update = function () {
+  IpRecord.prototype.update = function (action) {
+    // Increment account status check and throttle if needed
+    if (isAccountStatusAction(action)) {
+      this.addAccountStatusCheck()
+      if (this.isOverAccountStatusCheck() && !this.shouldBlock()){
```
rfk: Do we need to check `!this.shouldBlock()` here?  If it's only to guard against re-rate-limiting someone that's already limited, it may be clearer to directly call `!this.isRateLimited()`.

ip_record.js
```diff
@@ -3,36 +3,89 @@
...
+        this.rateLimit()
+      }
+    }
+
     return this.retryAfter()
```
rfk: It looks like `retryAfter` will return zero if the IP is not blocked or rate-limited, but the code doesn't make it obvious that this is an important behaviour.  It may be clearer to follow the pattern in e.g. `ip_email_record.js` where we explicitly `return 0` in the case that the user is not blocked.

test/remote/too_many_account_status_check.js
```diff
@@ -0,0 +1,136 @@
...
+
+Promise.promisifyAll(client)
+
+function delay(ms) {
+  var deferred = Promise.pending()
+  setTimeout(function(){
+    deferred.resolve()
+  }, ms)
+  return deferred.promise
+}
```
rfk: We should be able to use [1] directly here.

[1] http://bluebirdjs.com/docs/api/delay.html

test/local/ip_record_tests.js
```diff
@@ -10,7 +10,7 @@ function now() {
 }
 
 function simpleIpRecord() {
-  return new (ipRecord(120 * 1000, now))()
+  return new (ipRecord(120 * 1000, now, 90, 1))()
 }
```
rfk: I appreciate the thorough test coverage on the API side, but we should probably also add more local tests in this file to cover the new behaviors.

rfk:
@vbudhram good stuff!  Modulo the api-endpoint/action discussion we had today, this nicely matches what I had in mind.

One final design question to ponder: is there a better name than "account status check" for the action here?  I wonder if calling it something like "account lookup" or "email lookup" would help make it obvious precisely what is being limited and why.

rfk:
> Merge remote-tracking branch 'origin/master' into issue-96-add-accoun

FWIW, I find it's usually better to rebase PRs on top of master rather than merge with master, like:

```
git checkout master
git pull
git checkout my-pr-branch
git rebase master
git push -f
```

This re-creates the PR commits on top of whatever new stuff is in master, and avoid introducing merge commits into the PR.

README.md
```diff
@@ -41,13 +41,13 @@ On Mac OS X, memcached must be manually started for the tests to run.
...
-- `./lib/bans/`: code implementing temporary bans of specific email or IP addresses and listening on the SQS API for requests
-- `./lib/config/config.js`: where all of the configuration options are defined
-- `./lib/email_record.js`, `./lib/ip_email_record.js` and `./lib/ip_record.js`: code implementing the various blocking and rate-limiting policies
-- `./scripts`: helper scripts only used for development/testing
-- `./test/local`: unit tests
-- `./test/remote`: tests exercising the HTTP API
+- `bans/`: code implementing temporary bans of specific email or IP addresses and listening on the SQS API for requests
+- `bin/customs_server.js`: process listening on the network and responding to HTTP API calls
+- `config/config.js`: where all of the configuration options are defined
+- `email_record.js`, `ip_email_record.js` and `ip_record.js`: code implementing the various blocking and rate-limiting policies
```
rfk: Looks like you lost the "./" => "./lib/" rename on some of these file references.

docs/api.md
```diff
@@ -28,6 +28,7 @@ content-type of "application/json".
 * [POST /check](#post-check)
 * [POST /failedLoginAttempt](#post-failedloginattempt)
 * [POST /passwordReset](#post-passwordreset)
+* [POST /accountStatus](#post-accountstatus)
```
rfk: IIUC this new API endpoint no longer exists.

lib/ip_record.js
```diff
@@ -3,37 +3,96 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 // Keep track of events related to just IP addresses
-module.exports = function (BLOCK_INTERVAL_MS, now) {
+module.exports = function (BLOCK_INTERVAL_MS, RATE_LIMIT_INTERVAL_MS, MAX_ACCOUNT_STATUS_CHECK, now) {
 
   now = now || Date.now
 
+  var ACCOUNT_STATUS_ACTION = {
+    requestAccountStatus       : true
```
rfk: naming nit: none of the other actions have a prefix like "request" on them, something like "accountStatus" or "accountStatusCheck" would be more consistent with existing actions like "passwordForgotSendCode" and "accountCreate".

lib/ip_record.js
```diff
@@ -3,37 +3,96 @@
...
+  function isAccountStatusAction(action) {
+    return ACCOUNT_STATUS_ACTION[action]
+  }
+
   IpRecord.parse = function (object) {
     var rec = new IpRecord()
     object = object || {}
     rec.bk = object.bk // timestamp when the account was blocked
+    rec.as = object.as || []       // timestamps when account status check occurred
+    rec.rl = object.rl       // timestamp when the account was rate-limited
```
rfk: nit: comment indentation is inconsistent with the preceeding line; I don't have any particular opinion of the *correct* amount of indentation, but I do care about consistency :-)

lib/ip_record.js
```diff
@@ -3,37 +3,96 @@
...
+      }
+    }
+
+    // if an action is specified, return retry attempt
+    // which could
+    if (action) {
+      return this.retryAfter()
+    } else {
+      return 0
+    }
```
rfk: Hmm...I know I suggested it, but seeing it in the flesh, I'm not sure this explicit conditional and `return 0` is adding anything.  Will `action` ever be falsey?  I'd be happy for this to return to how you previously had it if you think it's simpler.

rfk:
@vbudhram this looks great to me!  I left a couple of small nits which you're welcome to ignore if they don't seem valuable to you.  r+

rfk:
(pending travis of course)

README.md
vbudhram: :/ Must have missed on merge, fixed

lib/ip_record.js
vbudhram: :+1:

lib/ip_record.js
vbudhram: :+1:

PR #100, opened 2016-02-19
fix(config): restore top-level "config" dir for $(NODE_ENV).json files.

rfk:
This was lost in #98 and broke new fxa-dev deployments.


rfk:
Trivial config fix that's breaking deploys, r=me

PR #107, opened 2016-04-12
fix(check): Include more action names in various checks.

rfk:
This ensures we treat all actions with similar affects as equivalent and rate-limit them together.  It's especially relevant for the many different ways that one can discover whether a particular account exists or not.

Note that the rate-limiting for account status check now counts the number of unique emails accessed within the rate-limiting window, not the total number of checks performed.  This should help mitigate the risk of adding all these other actions into that category.  We may still find we need to raise the limit on that though.

### DIFF
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -2,18 +2,34 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+// Actions that, if allowed, would allow an attacker
+// to try a candidtate password against an account.
 var PASSWORD_CHECKING_ACTION = {
   accountLogin: true,
   accountDestroy: true,
   passwordChange: true,
 }
 
+// Actions that, if allowed, would allow an attacker
+// to check whether an account exists for a particular user.
+// Basically any unauthenticated endpoint that takes
+// an email address as input.
 var ACCOUNT_STATUS_ACTION = {
+  accountCreate: true,
+  accountLogin: true,
+  accountDestroy: true,
+  accountLock: true,
+  accountUnlockResendCode: true,
+  passwordChange: true,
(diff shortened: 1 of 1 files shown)


rfk:
@vbudhram r?; also what do you think about raising the status-check limit here?  We may need to accompany this with better metrics in case if proves too restrictive in practice.

vbudhram:
Code looks good, I do agree on increasing the rate limiting, but I am unsure to what amount? It seems like this could drastically increase our rating limiting for normal users, or users sharing a network.

A scenario I could see is a user attempts to login a few times, gets rate limited, and is unable to reset their password because of this rate limit. But maybe the fix for that is to say that the password reset was sent?

rfk:
> A scenario I could see is a user attempts to login a few times, gets rate limited,
> and is unable to reset their password because of this rate limit.

Hmmm..yeah such a user is likely to have a bad time.  Two things I think we can do to help:

* Ensure that the (ip,email) limit on the number of password-checking actions, is more strict than the ip-based limit on the status-checking actions.  They way they get locked our of login attempts before they get locked out of email-sends.
* Allow them the option of a captcha to bypass the rate-limiting and get the email sent.

rfk:
I opened https://github.com/mozilla-services/puppet-config/pull/1941 to increase the limit here, it's currently lower than the limit on max-bad-logins which doesn't make a lot of sense.

rfk:
Also given that the counting is now done per unique email address, rather than counting all attempts, I think this can safely ship.  But I'd still like us to have better metrics on the things we're blocking and why...

vbudhram:
> Ensure that the (ip,email) limit on the number of password-checking actions, is more strict than the ip-based limit on the status-checking actions.

I do like adding this, but definitely not in the scope of this PR. 👍

rfk:
The puppet-config change landed, so I'm going to merge this for the next train.

PR #109, opened 2016-04-18
some *sharp* tools for updating the configs in memcache

dannycoates:
@jrgm is this good enough or do you want more fancy?


rfk:
Sharp indeed, but I think there's value in getting this out the door in the next release, and continuing to iterate as follow-ups if we need to.  r+ on the code side from me, will sync up with @jrgm for feedback before merging.

jrgm:
LGTM

jrgm:
One note: `chmod +x ./scripts/config-set.js`

dannycoates:
Thanks @jrgm!

> chmod +x ./scripts/config-set.js

fixed

PR #112, opened 2016-05-03
feat(email): Add config option to avoid blocking certain email domains

rfk:
Similar to our existing `ALLOWED_IPS` setting, this PR gives us the ability to allowlist a set of email address domains, exempting them from rate-limits and blocks.  This is likely very handy for QA, both for ourselves (where we currently have to raise rate limits in stage) and for others (e.g. https://bugzilla.mozilla.org/show_bug.cgi?id=1269425)

A couple of notes:

* The `allowed_email_domains.js` file is a copy-paste of the existing `allowed_ips.js` file; I considered a refactor to extract shared logic but decided not to invest in it at this time
* The checking of the allowlist is only done at the point where we are actually about to block an email, which means there should be no additional overhead for legitimate users

@seanmonstar r?

### DIFF
--- a/lib/allowed_email_domains.js
+++ b/lib/allowed_email_domains.js
@@ -1,0 +1,67 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports = function (config, mc, log) {
+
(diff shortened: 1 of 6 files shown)


rfk:
Also, this subtly widens the scope of the `allowed_ips` setting - previously it would exempt you from any ip-specific rate-limiting, but now it exempts you from all blocks, even those based on email address.  I think that's in line with the intended semantics of these allowlists, but am open to suggestion that it's overly broad.

PR #114, opened 2016-05-24
feat(recaptcha): support entry reset

vladikoff:
Connects to https://github.com/mozilla/fxa/issues/154

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -242,6 +242,49 @@ module.exports = function createServer(config, log) {
   )
 
   api.post(
+    '/resetEntry',
+    function (req, res, next) {
+      var email = req.body.email
+      var ip = req.body.ip
+
+      if (!email || !ip) {
+        var err = {code: 'MissingParameters', message: 'email, ip are all required'}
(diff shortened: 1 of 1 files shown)


lib/server.js
```diff
@@ -242,6 +242,49 @@ module.exports = function createServer(config, log) {
...
+      fetchRecords(email, ip)
+        .spread(
+          function (emailRecord, ipRecord, ipEmailRecord) {
+            emailRecord.lf = []
+            ipRecord.lf = []
+            ipEmailRecord.lf = []
+
+            ipRecord.rl = null
+            ipEmailRecord.rl = null
+            emailRecord.rl = null
```
rfk: IIUC, this is clearing just the login failure information.  Is that because we only show the captcha on sign-in views for now?

There are other reasons that the user might get an "attempt limit exceeed" error, e.g. because of sending too many emails, because of being explicitly blocked at the IP level, or (soon) because their IP is on some third-party blocklist.
We need to be careful that we don't give a "limit exceeded" error, throw up a captcha for the user, them block them with a "limit exceeded" error again anyway.

PR #117, opened 2016-06-21
feat(blocklist): Add blocklist module

vbudhram:
This PR adds the ability to check a blocklist of ips and automatically block that request. Ref: https://github.com/mozilla/fxa/pull/156, https://github.com/mozilla/fxa-features/issues/6

Unanswered questions:

- [x] Retrieve valid Spamhaus BCL file
- [ ] ~~Determine path forward for retrieving updates of file~~
- [ ] ~~Create standalone module or service~~
- [x] Add metrics logging
- [x] Add feature flag and logOnly mode
- [x] Add automatic reload on blocklist file change
- [x] Add support for multiple blocklist
- [x] Add metrics for time taken to process
- [ ] ~~Add ability to update list~~ https://github.com/mozilla/fxa-customs-server/issues/130
- [x] Add firehol lists


lib/server.js
```diff
@@ -31,6 +31,8 @@ module.exports = function createServer(config, log) {
   var limits = require('./limits')(config, mc, log)
   var allowedIPs = require('./allowed_ips')(config, mc, log)
   var allowedEmailDomains = require('./allowed_email_domains')(config, mc, log)
+  var blocklist = require('./blocklist')
+  blocklist.load(config.blocklist.listPath)
```
vladikoff: do we need to log of successful here or catch if loading failed here?

package.json
```diff
@@ -23,6 +23,7 @@
     "bluebird": "3.3.4",
     "bunyan": "1.8.0",
     "convict": "1.2.0",
+    "csv-parser": "^1.9.3",
```
vladikoff: remove the `^`

lib/blocklist.js
```diff
@@ -0,0 +1,48 @@
+/* Any copyright is dedicated to the Public Domain.
```
vladikoff: Should ths be MPL instead of Public Domain here?

lib/blocklist.js
```diff
@@ -0,0 +1,48 @@
...
+  },
+  clear: function () {
+    blocklist = undefined
+  },
+  contains: function (ip) {
+    if (!blocklist) {
+      // Throw error if blocklist has not been loaded
+      throw new Error('Please load blocklist')
+    } else {
+      return blocklist[ip] ? true : false
```
vladikoff: nit: I think we try to avoid ternary operators in our code base

lib/blocklist.js
```diff
@@ -0,0 +1,48 @@
...
+          ips[row] = {}
+        })
+        blocklist = ips
+      })
+  },
+  clear: function () {
+    blocklist = undefined
+  },
+  contains: function (ip) {
+    if (!blocklist) {
```
vladikoff: nit: needs a space after `! ` (content server rules?)

lib/config/config.js
```diff
@@ -159,7 +159,15 @@ module.exports = function (fs, path, url, convict) {
         'restmail.net',
         'mockmyid.com'
       ]
-    }
+    },
+    blocklist: {
+      listPath: {
+        doc: 'Location of the csv file containing list of ip addresses to block',
```
vladikoff: does this need to state that this relative to the fxa-customs-server dir? (because it uses ` csvPath = path.join(__dirname, csvPath)` ?

lib/blocklist.js
```diff
@@ -0,0 +1,48 @@
...
+      // Throw error if blocklist has not been loaded
+      throw new Error('Please load blocklist')
```
vladikoff: Might need to adjust this. Is throwing here risky in terms of breaking a lot of requests? Should the blocklist component be more isolated (if it is not loaded, log the fatal / error but proceed with the customs request)?

lib/blocklist.js
vbudhram: I'm not sure, some files have Public Domain, some have MPL, @rfk do we have a stance on which?

vladikoff: From mozilla-central: the tests are public domain, other stuff is MPL.

I think Public domain for tests is because you can take the tests and run against your own implementation of something (maybe?)

rfk: General policy is MPL by default, unless we have a specific reason to use a different license.

vbudhram:
@seanmonstar Mind giving this a look over?

lib/server.js
```diff
@@ -47,6 +47,19 @@ module.exports = function createServer(config, log) {
 
   var handleBan = P.promisify(require('./bans/handler')(config.memcache.recordLifetimeSeconds, mc, EmailRecord, IpRecord, log))
 
+  // Setup blocklist
+  if (config.ipBlocklist.enable) {
+    var IPBlocklist = require('./ip_blocklist')(log)
+    var ipBlocklist = new IPBlocklist()
+    ipBlocklist.load(config.ipBlocklist.listPath)
```
seanmonstar: It seems unlikely, but could this load hang longer than expected, and so no be ready when requests are serving?

lib/server.js
```diff
@@ -111,6 +124,17 @@ module.exports = function createServer(config, log) {
       }
       email = normalizedEmail(email)
 
+      // Check blocklist before retrieving records, save a lookup
+      if (config.ipBlocklist.enable && ipBlocklist.contains(ip)) {
+        log.debug({op: 'ipBlocklist.hit', ip: ip, action: action, list: config.listPath})
+
+        if (!config.logOnly) {
```
seanmonstar: Probably needs to be `config.ipBlocklist.logOnly`

test/remote/ip_blocklist_tests.js
```diff
@@ -0,0 +1,97 @@
...
+  function (t) {
+    client.postAsync('/check', {ip: VALID_IP, email: TEST_EMAIL, action: ACTION},
+      function (err, req, res, obj) {
+        t.equal(obj.block, false, 'request is not blocked')
+        t.end()
+      }
+    )
+  }
+)
+
```
seanmonstar: A test for `logOnly` would catch the `config.logOnly` typo.

lib/ip_blocklist.js
```diff
@@ -0,0 +1,96 @@
...
+  IPBlocklist.prototype.contains = function (ipAddress) {
+    var self = this
+
+    try {
+      return self.prefixLengths.some(function (prefixLength) {
+        return getBaseIp(ipAddress, prefixLength) in self.ipsByPrefixLength[prefixLength]
+      })
+    } catch(err){
+      // On any error, side on caution and allow the request through. These errors could be
+      // malformed ip address, etc
```
rfk: In other places in customs-server, we deliberately fail closed for security.  It's probably worth being consistent here.

vbudhram:
@seanmonstar Made updates you recommended. Biggest change is now customs server waits for blocklist to be loaded before allowing requests through. Mind another r?

seanmonstar:
r+ with squash

vladikoff:
@seanmonstar you can use the new GitHub squash button to merge via:

vbudhram:
@seanmonstar Thanks for review! I think we should block this pending my 1:1, want to ensure correct path forward for it.

vbudhram:
Spoke with @rfk and we decided that this feature should have the ability to reload the blocklist if it detects a change to the file. Doing this will defer the fetching and updating of the list to ops.

vbudhram:
Blocked on getting lists to test.

vbudhram:
@rfk @seanmonstar I think this ready for another review? Mind giving it a look over? 

Should I include the firehol ipsets?

lib/config/config.js
```diff
@@ -167,6 +167,39 @@ module.exports = function (fs, path, url, convict) {
...
+        doc: 'Flag to only log hits',
+        format: Boolean,
+        default: true,
+        env: 'IP_BLOCKLIST_LOGONLY'
+      },
+      lists: {
+        doc: 'A array of ip blocklist file paths',
+        format: Array,
+        default: [
+          './../test/mocks/firehol_level1.netset',
```
seanmonstar: Looks funny to have test mocks as the default in config. Seems they should either be in a test config, or the default config should have its files promoted out of 'mocks'.

lib/ip_blocklist_manager.js
```diff
@@ -0,0 +1,88 @@
...
+
+  IPBlocklistManager.prototype.contains = function (ipAddress) {
+    var self = this
+    var startTime = Date.now()
+    var endTime
+    var found = false
+
+    // Go through all blocklists and check if there is a match on the
+    // ipAddress.
+    self.ipBlocklists.forEach(function (blocklist) {
```
seanmonstar: `forEach` will continue the loop and call `blocklist.contains` on every list even once a match is found. I'd use [`Array.some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) so that once a block is found, the lookup can stop right away.

```js
var found = self.ipBlocklists.some(function(list) { return list.contains(ipAddress) })
```

lib/ip_blocklist_manager.js
```diff
@@ -0,0 +1,88 @@
...
+        found = true
+      }
+    })
+
+    if (found) {
+      endTime = Date.now()
+      log.info({
+        op: 'IPBlocklistManager.contains',
+        ip: ipAddress,
+        fileName: self.fileName,
```
seanmonstar: Will this include the correct `fileName`? It reads like the `IPBlocklistManager` actually contains many lists coming from many files. I also don't see `this.fileName` set anywhere in this module.

lib/config/config.js
vbudhram: Yea you are right, I can remove and make no lists the default.

lib/ip_blocklist_manager.js
vbudhram: Interestingly, this was the functionality I wanted. In this phase, we wanted to gather metrics on how well each list performs and not break loop on a hit. I will update comment to make this more explicit.

lib/ip_blocklist_manager.js
vbudhram: Good catch, will remove.

lib/ip_blocklist_manager.js
rfk: In my head, I can see it being useful to check all the lists even if we have found a match in one of them.  It will be useful for ongoing metrics purposes, e.g. if the block later proves to have been a false-positive, it will be useful to log that as a metric against *all* the security features that matched it, not just the first one we found.

seanmonstar: @rfk I can see that being useful. It's not currently in this code, so it seems like unnecessary cpu cycles. To do what you suggest, perhaps an array should be built of list names that contain the address?

vladikoff:
from mtg: needs final r

vbudhram:
@seanmonstar Thanks for feedback, made updates! @rfk Mind a look over?

lib/ip_blocklist.js
```diff
@@ -0,0 +1,165 @@
...
+
+    try {
+      return self.prefixLengths.some(function (prefixLength) {
+        var result = getBaseIp(ipAddress, prefixLength) in self.ipsByPrefixLength[prefixLength]
+
+        // If `hit` log metrics on time taken
+        if (result) {
+          endTime = Date.now()
+          log.info({op: 'fxa.customs.blocklist.hit', ip: ipAddress, fileName: self.fileName, fileSize: self.fileSize,
+            foundIn: (endTime - startTime)})
```
rfk: nit: the indentation style here doesn't match our usual convention, which IIRC would expand this to have each item on its own line:

```
lof.info({
  op: ...,
  ip: ...,
  ...
})
```

lib/ip_blocklist.js
```diff
@@ -0,0 +1,165 @@
...
+    var self = this
+
+    log.trace({op: 'fxa.customs.blocklist.refreshList', fileName: self.fileName, fileSize: self.fileSize})
+
+    return statFile(self.filePath)
+      .then(function (fileStats) {
+        var mtime = fileStats.mtime
+        if (mtime > self.fileLastModified) {
+          self.fileLastModified = mtime
+          return self.load(self.filePath)
```
rfk: So IIUC, this causes the list to be reloaded in-place.  Will it successfully remove any ips that were previously in the list, but have since been removed from it?  From a quick read it looks like it might only add new entries without removing the old ones.

Another concern is, how long does the reload typically take?  We need to be careful that we don't block the entire event loop while reloading a big file.

lib/ip_blocklist.js
```diff
@@ -0,0 +1,165 @@
...
+          var parsedData = parseRow(row[0])
+
+          if (parsedData) {
+            if (!self.ipsByPrefixLength[parsedData.prefixLength]) {
+              self.prefixLengths.push(parsedData.prefixLength)
+              self.ipsByPrefixLength[parsedData.prefixLength] = {}
+            }
+            self.ipsByPrefixLength[parsedData.prefixLength][parsedData.baseIp] = true
+          }
+        })
```
rfk: One simple option here to deal with reload of the list, might be to make fresh local instances of the state variables:

```
var newIpsByPrefixLength = {}
var newPrefixLenghts = []
```

Load the data from the file into these new objects, then at the end swap them in for the old state:

```
self.ipsByPrefixLength = newIpsByPrefixLength
self.prefixLenghts = newPrefixLenghts
```

lib/ip_blocklist.js
```diff
@@ -0,0 +1,165 @@
...
+      .then(function (fileStats) {
+        self.fileLastModified = fileStats.mtime
+        self.fileSize = fileStats.size
+        return readFile(filePath, 'utf8')
+      })
+      .then(function (data) {
+        return parse(data)
+      })
+      .then(function (rows) {
+        rows.forEach(function (row) {
```
rfk: If there are a lot of rows here, this will block the event loop until we're processed each row.  There may be an async iteration helper in bluebird that can avoid this by processing them in small batches.

lib/server.js
```diff
@@ -112,6 +121,15 @@ module.exports = function createServer(config, log) {
       }
       email = normalizedEmail(email)
 
+      // Check blocklist before retrieving records, save a lookup
```
rfk: While saving lookups is usually a good idea, this will bypass the logic that checks for `allowedIPs` and `allowedEmails`, as well as bypassing some metrics logging.

I think we'll need to move this somewhere inside the `fetchRecords` handler, and/or move the `allowedIPs` logic outside of that handler; the private repo has a bit of a refactor of this code that may help here, see https://github.com/mozilla/fxa-customs-server/issues/126

My general philosophy for this codebase, is to optimize for returning `block: false` as quickly as possible, but it's OK to do extra work for requests that we're going to block.  So doing an unnecessary `fetchRecords` for IPs that are on a blocklist, is OK by me.

lib/ip_blocklist.js
vbudhram: I think from doing https://github.com/mozilla/fxa-customs-server/pull/117#discussion_r79723749, it will ensure a clean list is loaded. 

> Another concern is, how long does the reload typically take?

Reloading on a 1mb file typically takes 300ms on my computer and will only occur if a change in the file is detected. A scenario I could see blocking event loop significantly is if all blocklist files were updated, then they could all try to reload themselves at the same time. 

Not sure if that is best handled in here or in the updating script?

lib/server.js
vbudhram: > this will bypass the logic that checks for allowedIPs and allowedEmails

Good point! I had not thought about that. Will revise the logic for this.

lib/ip_blocklist.js
vbudhram: I took a guess of 10k for batch sizes with delay of 100ms between. Seem reasonable?

rfk: > delay of 100ms between

I don't think you even need a delay, you just need to yield to the event loop.  [setImmediate](https://nodejs.org/api/all.html#all_setimmediate_callback_arg) or similar is likely what you're after (or better yet, some abstracted-away looping equivalent)

lib/ip_blocklist.js
rfk: > Reloading on a 1mb file typically takes 300ms on my computer
> [...]
> I took a guess of 10k for batch sizes 

Assuming 16 bytes per line, that's 65k lines in 300ms, or about 45ms per 10k lines.  That's still a long time not to do any I/O.  1k batches maybe?  I dunno, it's all guesswork at this point :-)

lib/ip_blocklist.js
rfk: > (or better yet, some abstracted-away looping equivalent)

I thought bluebird had such a thing, but I can't see it in their API documentation.  One way to hack it might be something like:

```
Promise.each(rows, function(row, idx) {
  ...do stuff with the row...
  if (idx % 1000 === 0) {
    return P.fromNode(setImmediate)
  }
})
```

lib/config/config.js
```diff
@@ -167,6 +167,44 @@ module.exports = function (fs, path, url, convict) {
...
+        doc: 'Poll interval for checking for updated lists (seconds)',
+        default: 300,
+        format: 'nat',
+        env: 'IP_BLOCKLIST_POLL_INTERVAL_SECONDS'
+      },
+      batchSize: {
+        doc: 'Batch size to process lists',
+        default: 10000,
+        format: 'nat',
+        env: 'IP_BLOCKLIST_BATCH_SIZE'
```
rfk: TBH I don't think you even need to make this configurable, just pick a sensible default in the file itself.

test/local/ip_blocklist_tests.js
```diff
@@ -0,0 +1,207 @@
...
+
+  test(
+    name + ', returns true for invalid ip',
+    function (t) {
+      var ipBlocklist = new BlocklistClass()
+
+      ipBlocklist.load(filePath)
+        .then(function () {
+          var result = ipBlocklist.contains('notip')
+          t.equal(result, true, 'return true for invalid ip')
```
rfk: Hrm...is returning `true` when we're given invalid input the right thing to do?  We certainly want to block the request if the ip is missing or invalid, but that may be better handled by an explicit validation in server.js rather than assuming this functionality from the blocklist.

test/local/ip_blocklist_tests.js
```diff
@@ -0,0 +1,207 @@
...
+      })
+      .then(function () {
+        t.end()
+      })
+      .catch(function (err) {
+        t.fail(err)
+        t.end()
+      })
+  }
+)
```
rfk: It would be interesting to turn this into an explicit test that we correctly handle both additions and deletions from the list.  One option would be to:

* make a tempfile containing an initial version of a list
* load it and test that it works correctly
* overwrite the tempfile with new contents, removing some ips and adding others
* refresh it, and make sure it reflects the expected changes

Not sure how much work that would be to actually write in practice though :-)

test/remote/ip_blocklist_tests.js
```diff
@@ -0,0 +1,106 @@
...
+var VALID_IP = '3.0.0.0'
+
+process.env.IP_BLOCKLIST_ENABLE = true
+process.env.IP_BLOCKLIST_LOGONLY = false
+process.env.IP_BLOCKLIST_FILES = [
+  './../test/mocks/firehol_level1.netset',
+  './../test/mocks/firehol_level2.netset',
+  './../test/mocks/firehol_level3.netset',
+  './../test/mocks/firehol_level4.netset',
+  './../test/mocks/firehol_webserver.netset',
```
rfk: I'm not sure there's much value in committing *all* of these files to the repo here, especially the big ones.  Would it suffice to include only the leve1 list, for syntax-compatibility-testing purposes?

rfk:
@vbudhram I feel like I left a lot of comments, but overall this is looking in *really* good shape!

lib/ip_blocklist.js
rfk: (Actually I seem remember thinking that `Promise.each` would give you this behaviour for free, but I can't find any proof that that's the case)

test/local/ip_blocklist_tests.js
vbudhram: I think this is fair, the default is to block on [server-side errors](https://github.com/mozilla/fxa-customs-server/blob/200d7e97a88a478430ea59723886d48786205b00/lib/server.js#L186). I can rethrow the error and have the server handle it.

vbudhram:
@rfk Thanks for feedback! All things considered, I don't think it needs to hold up tagging this train. There is still a little work that needs to be done to get this up to par, but nothing to crazy. Will get these in soon!

vbudhram:
@rfk Made the updates you recommended, sending back your way!

lib/ip_blocklist.js
```diff
@@ -0,0 +1,196 @@
...
+      op: 'fxa.customs.blocklist.refreshList',
+      fileName: self.fileName,
+      fileSize: self.fileSize
+    })
+
+    return statFile(self.filePath)
+      .then(function (fileStats) {
+        var mtime = fileStats.mtime
+        if (mtime > self.fileLastModified) {
+          self.fileLastModified = mtime
```
rfk: IIUC the call to `self.load` will update `self.fileLastModified` for us, so I don't think we need to do it again here?

lib/server.js
```diff
@@ -136,15 +145,29 @@ module.exports = function createServer(config, log) {
 
             var retryAfter = [blockEmail, blockIpEmail, blockIp].reduce(max)
 
+            var allowIP = ip in allowedIPs.ips
+            var allowedEmailDomain = allowedEmailDomains.isAllowed(email)
```
rfk: naming nit: `allowedIP` rather than `allowIP` for consistency with the other names in this block.

lib/server.js
```diff
@@ -136,15 +145,29 @@ module.exports = function createServer(config, log) {
...
+            // IP's that are in blocklist should be blocked
+            // and not return a retryAfter because it is not known
+            // when they would be removed from blocklist
+            if (config.ipBlocklist.enable && blockListManager.contains(ip) && !allowIP && !allowedEmailDomain) {
+              if (!config.ipBlocklist.logOnly) {
+                return {
+                  block: true
+                }
+              }
+            }
```
rfk: Returning here means we won't call `setRecords`, so we won't write back any updated memcached records for IP addresses on the blocklist.  Your call whether that's important or not, just want to note it explicitly.

test/local/ip_blocklist_tests.js
```diff
@@ -0,0 +1,279 @@
...
+      })
+      .then(function () {
+        var result = ipBlocklist.contains(tmpFileContents)
+        t.equal(result, false, 'should not contain ip')
+
+        result = ipBlocklist.contains(tmpFileContents2)
+        t.equal(result, true, 'should contain ip')
+
+        t.end()
+      })
```
rfk: nice!

test/remote/ip_blocklist_logonly.js
```diff
@@ -0,0 +1,72 @@
...
+})
+
+Promise.promisifyAll(client, {multiArgs: true})
+
+test(
+  'log only from blocklist',
+  function (t) {
+    client.postAsync('/check', {ip: BLOCK_IP, email: TEST_EMAIL, action: ACTION},
+      function (err, req, res, obj) {
+        t.equal(obj.block, false, 'request is blocked')
```
rfk: "request is blocked" -> "request is not blocked".  This also doesn't test whether we actually logged anything, although if we aren't set up to test for logging output in this repo, that may be a significant chunk of work that can be deferred to a follow-up issue.

test/remote/ip_blocklist_tests.js
vbudhram: This was added here https://github.com/mozilla/fxa-customs-server/blob/1a7661cb28d4139d5b00a23de73e77bb35de80e5/test/remote/ip_blocklist_logonly.js

vbudhram:
@seanmonstar Made changes you recommended, you approve?

vbudhram:
Thanks for reviews! @rfk @seanmonstar

vladikoff:
Awesome!

rfk:
@vbudhram nice work pushing this through!  I feel like there's a follow-up bug for Ops to discuss adding the necessary files in deployment, can you please create one or link to an existing one from here?  I guess https://github.com/mozilla-services/puppet-config is the right place for it either way.

vbudhram:
@rfk I think we should be able to enable this feature on next train. It doesn't look like I have access to puppet config repo (possible to get access?). I created one a little while back to track the next steps, issue https://github.com/mozilla/fxa-customs-server/issues/130.

rfk:
I've invited you to puppet-config with read access; just remember not to fork it anywhere public :-)

W.r.t. next steps, I feel like there's an intermediate step between here and #130 - namely, having the build process bake in snapshots of the lists at build time, in the same way that we currently handle the maxmind database.  This will let us ship relatively-up-to-date versions of several different lists, without having to include them in the customs-server repo itself, and without having to solve the entire automagical-update problem.

PR #119, opened 2016-07-20
fix(deps): update dev dependencies

vladikoff:
@vbudhram r?


package.json
```diff
@@ -29,16 +29,16 @@
   },
   "devDependencies": {
     "ass": "git://github.com/jrgm/ass.git#5be99ee7abc9fcf63f9ebcc37b151b9c822146d1",
```
vbudhram: General question, noticed we removed `ass` from other services. Should we do a follow up issue to do same?

vbudhram:
LGTM 👍

package.json
vladikoff: Yeap that's the plan! 👍

PR #121, opened 2016-07-22
feat(server): Add Uid_Record and checkAuthenticated endpoint

larissagaulia:


lib/config/config.js
```diff
@@ -94,6 +94,12 @@ module.exports = function (fs, path, url, convict) {
         format: 'nat',
         env: 'IP_RATE_LIMIT_BAN_DURATION_SECONDS'
       },
+      uidRateLimitBanDurationSeconds: {
+        doc: 'Duration of automatic ban',
+        default: 60 * 15,
+        format: 'nat',
+        env: 'UID_RATE_LIMIT_BAN_DURATION_SECONDS'
+      },
```
vladikoff: @larissagaulia, need to add # of attempts allowed here?

rfk:
@larissagaulia this is a great start!  I had a promised @vladikoff I'd file an issue on this repo with some suggestions for how to do uid-based ratelimiting, but you beat me to it...so instead I'll pile on here with some high-level comments :grin:

I like the way you've added a `/checkAuthenticated` endpoint to separate these out from the existing unauthenticated checks.  I think it would be worth paralleling the `/check` endpoint even more closely, e.g. having it take an ip, uid, and action like:

```
POST /checkAuthenticated
{
  ip: '1.2.3.4'
  uid: 'abcdef1234567890',
  action: 'uploadAvatar'
}
```

This will let us re-use that endpoint for rate-limiting authenticated actions in several different places.  Inside the UidRecord, you could keep a list of the timestamps at which various actions were performed by that uid.  For example a record might look like:

```
{
  "blockedTimestamp": null   // time at which this uid was blocked, or null if not blocked
  "actions": {                           // maps actions to timestamps at which they were performed,
      "avatarUpload": [ 12345, 12456],    // two "avatarUpload" actions performed recently
      "devicesNotify": [12567]                  // one "devicesNotify" action performed recently
   }
}
```

Then, when checking whether an action can be performed, look in the list to see how many were performed recently, and block if there were too many. 

You could store a similar set of authenticated actions in the IpRecord to rate-limit based on IP as well as on uid.

lib/server.js
```diff
@@ -164,6 +184,59 @@ module.exports = function createServer(config, log) {
   )
 
   api.post(
+    '/checkAuthenticated',
+    function (req, res, next) {
+      var ip = req.body.ip
+      var token = req.body.token
```
rfk: If you don't need the token for anything, don't pass it to customs-server here - the less often we send tokens over the wire, the less likely we are to accidentally leak them somewhere they shouldn't go.

lib/config/config.js
larissagaulia: @vladikoff, with 15 min, how many attempts you think reasonable? 50?

vladikoff: Let's go with a 100 for now 👍

lib/config/config.js
```diff
@@ -94,6 +94,12 @@ module.exports = function (fs, path, url, convict) {
         format: 'nat',
         env: 'IP_RATE_LIMIT_BAN_DURATION_SECONDS'
       },
+      uidRateLimitBanDurationSeconds: {
```
vladikoff: @larissagaulia for these new values, do you think it will be cleaner to nest the values like here: https://github.com/mozilla/fxa-auth-server/blob/master/config/index.js#L365-L386

so you would have a top level `uidRateLimit` and then nested in it `uidRateLimit.banDurationSeconds`, etc

larissagaulia: like it

lib/config/config.js
```diff
@@ -94,12 +94,24 @@ module.exports = function (fs, path, url, convict) {
...
         default: 60 * 15,
         format: 'nat',
         env: 'UID_RATE_LIMIT_BAN_DURATION_SECONDS'
       },
+      maxChecksPerUid: {
+        doc: 'Number of checks within uidRateLimitBanDurationSeconds before blocking',
+        default: 100,
+        format: 'nat',
+        env: 'MAX_CHECKS_PER_UID'
+      },
```
vladikoff: @larissagaulia right so for above you can do: 

```
uidRateLimit: {
      limitIntervalSeconds: {
        doc: 'Duration of automatic throttling for uids',
        default: 60 * 15,
        format: 'nat',
        env: 'UID_RATE_LIMIT_INTERVAL_SECONDS'
      },
      banDurationSeconds: {
        doc: 'Duration of automatic ban',
        default: 60 * 15,
        format: 'nat',
        env: 'UID_RATE_LIMIT_BAN_DURATION_SECONDS'
      },
      maxChecks: {
        doc: 'Number of checks within uidRateLimitBanDurationSeconds before blocking',
        default: 100,
        format: 'nat',
        env: 'UID_RATE_LIMIT'
      }
}
```

`MAX_CHECKS_PER_UID` -> `UID_RATE_LIMIT`

lib/server.js
```diff
@@ -186,13 +186,14 @@ module.exports = function createServer(config, log) {
   api.post(
     '/checkAuthenticated',
     function (req, res, next) {
+      var action = req.body.action
       var ip = req.body.ip
       var token = req.body.token
```
vladikoff: @larissagaulia let's drop the `token` from everywhere

lib/config/config.js
larissagaulia: ohhh, You've given spoilers of my next push ='(
hahha

lib/server.js
```diff
@@ -164,6 +184,59 @@ module.exports = function createServer(config, log) {
...
+        return next()
+      }
+
+      fetchUidRecord(uid)
+        .spread(
+          function (uidRecord) {
+
+            var retryAfter = uidRecord.addCount(action, uid)
+
+            return setUidRecord(uid, uidRecord)
```
vladikoff: @larissagaulia we should also add support for `allowedEmailDomains`, means `@restmail.net` will not be limited.

```
if (allowedEmailDomains.isAllowed(email)) {
///
}
```

Ref: https://github.com/larissagaulia/fxa-customs-server/blob/87960655679a9fa909b2bdc325fc280765282144/lib/server.js#L151

test/remote/too_many_login_checks.js
```diff
@@ -158,6 +167,105 @@ test(
   }
 )
 
+test(
+  '/checkAuthenticated with same action',
```
vladikoff: @larissagaulia we should move these tests to their own file, `/checkAuthenticated` is not part of `test/remote/too_many_login_checks.js`

test/remote/too_many_login_checks.js
```diff
@@ -158,6 +167,105 @@ test(
...
+        t.end()
+      })
+  }
+)
+
+test(
+  '/checkAuthenticated with different actions',
+  function (t) {
+
+    // Send requests until one gests rate limited
```
vladikoff: `gests` -> `gets`

lib/uid_record.js
```diff
@@ -0,0 +1,96 @@
...
+
+    return this.getPossibleRetryDelay(action)
+  }
+
+  UidRecord.prototype.trimCheckAuthenticated = function (action, now) {
+    this.timestampsByAction[action] = this._trim(now, this.timestampsByAction[action], limits.maxChecksPerUid)
+  }
+
+  UidRecord.prototype._trim = function (now, items, max) {
+
```
vladikoff: removes this extra line break

lib/server.js
```diff
@@ -164,6 +184,64 @@ module.exports = function createServer(config, log) {
...
   api.post(
+    '/checkAuthenticated',
+    function (req, res, next) {
+      var action = req.body.action
+      var email = req.body.email
+      var ip = req.body.ip
+      var uid = req.body.uid
+
+      if(!action || !ip || !uid){
+        var err = {code: 'MissingParameters', message: 'action, email, ip and uid are all required'}
```
vladikoff: @rfk, I just realized that we cannot always provide the email here. How should we allow `@restmail` to be unblocked?

cc @larissagaulia

rfk: from mtg: let's not worry about that for now.  The limits here are a lot more generous than on e.g. number of login attempts.  So If we have some QA environment that's hammering us hard enough to trigger one of these rate-limits, that's a good signal that we should fix the QA environment rather than allow it to hammer harder.

larissagaulia: just to be clear: not worry about not rate limiting allowedEmailDomains, right? that is, let restmail be rate limited the same as all the others.

rfk: correct

lib/server.js
vladikoff: Nvm we are not doing that.

test/remote/too_many_authenticated_checks.js
```diff
@@ -0,0 +1,166 @@
...
+
+        return client.postAsync('/checkAuthenticated', { action: ACTION_B, ip: TEST_IP, uid: TEST_UID})
+      })
+
+      .spread(function(req, res, obj){
+        t.equal(res.statusCode, 200, 'returns a 200')
+        t.equal(obj.block, false, 'not rate limited for actionB after check2')
+
+        // Delay ~5s for rate limit to go away
+        return Promise.delay(5010)
```
vladikoff: nit: let's see if we can speed this up and have this be 2 seconds? (so tests run faster)

lib/server.js
```diff
@@ -164,6 +184,59 @@ module.exports = function createServer(config, log) {
...
+                    uid: uid,
+                    uidRecord: uidRecord
+                  }
+                }
+              )
+          }
+        )
+        .then(
+          function (result) {
+            log.info({ op: 'request.check', block: result.block })
```
vladikoff: @vladikoff need to check if this is the best metrics we can get and if it is enough.

vladikoff: `request.check'` -> 'request.checkAuthenticated',

lib/server.js
```diff
@@ -164,6 +184,59 @@ module.exports = function createServer(config, log) {
...
   api.post(
+    '/checkAuthenticated',
+    function (req, res, next) {
+      var action = req.body.action
+      var ip = req.body.ip
+      var uid = req.body.uid
+
+      if(!action || !ip || !uid){
+        var err = {code: 'MissingParameters', message: 'action, ip and uid are all required'}
+        log.error({op:'request.check', action: action, ip: ip, uid: uid, err: err})
```
vladikoff: `request.check'` -> 'request.checkAuthenticated',

lib/server.js
```diff
@@ -164,6 +184,59 @@ module.exports = function createServer(config, log) {
...
+              )
+          }
+        )
+        .then(
+          function (result) {
+            log.info({ op: 'request.check', block: result.block })
+            res.send(result)
+          },
+          function (err) {
+            log.error({ op: 'request.check', err: err })
```
vladikoff: `request.check'` -> 'request.checkAuthenticated',

lib/server.js
```diff
@@ -164,6 +165,60 @@ module.exports = function createServer(config, log) {
...
+
+            return setRecord(uid, uidRecord)
+              .then(
+                function () {
+
+                  return {
+                    block: retryAfter > 0,
+                    retryAfter: retryAfter,
+                    uid: uid,
+                    uidRecord: uidRecord
```
vladikoff: Final version should only have `block` and maybe `retryAfter` ?

lib/server.js
```diff
@@ -164,6 +165,60 @@ module.exports = function createServer(config, log) {
...
+        log.error({op:'request.checkAuthenticated', action: action, ip: ip, uid: uid, err: err})
+        res.send(400, err)
+        return next()
+      }
+
+      mc.getAsync(uid)
+        .then(UidRecord.parse, UidRecord.parse)
+        .then(
+          function (uidRecord) {
+
```
vladikoff: remote this line break

lib/server.js
```diff
@@ -164,6 +165,60 @@ module.exports = function createServer(config, log) {
...
+            var retryAfter = uidRecord.addCount(action, uid)
+
+            return setRecord(uid, uidRecord)
+              .then(
+                function () {
+
```
vladikoff: remove this line break

lib/server.js
```diff
@@ -164,6 +165,60 @@ module.exports = function createServer(config, log) {
...
+          }
+        )
+        .then(
+          function (result) {
+            log.info({ op: 'request.checkAuthenticated', block: result.block })
+            res.send(result)
+          },
+          function (err) {
+            log.error({ op: 'request.checkAuthenticated', err: err })
+
```
vladikoff: remove line break

lib/uid_record.js
```diff
@@ -0,0 +1,95 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// Keep track of events related to uid
+module.exports = function (limits, now) {
+
```
vladikoff: remove line break

lib/uid_record.js
```diff
@@ -0,0 +1,95 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// Keep track of events related to uid
```
vladikoff: format nit:

```
// Keep track of events related to uid
```

to
```
/**
This module keeps track of events related to a specific uid
*/

**Comment by vladikoff** [comment]
Needs rebase, passing travis and new docs in https://github.com/mozilla/fxa-customs-server/blob/master/docs/api.md#api-endpoints and after https://github.com/mozilla/fxa-customs-server/blob/master/docs/api.md#post-check

File: lib/config/config.js [hunk 21]
```diff
@@ -94,6 +94,26 @@ module.exports = function (fs, path, url, convict) {
...
+        },
+        banDurationSeconds: {
+          doc: 'Duration of automatic ban',
+          default: 60 * 15,
+          format: 'nat',
+          env: 'UID_RATE_LIMIT_BAN_DURATION_SECONDS'
+        },
+        maxChecks: {
+          doc: 'Number of checks within uidRateLimitBanDurationSeconds before blocking',
+          default: 100,
```
vladikoff: 👍

lib/uid_record.js
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  UidRecord.prototype.isRateLimited = function() {
+    //uid is rateLimited if there's a valid rateLimited timestamp and
+    //waiting time interval wasn't reached
+    return (this.rateLimitedTimestamp && (this.rateLimitedTimestamp + limits.uidRateLimitBanDurationMs > now()))
+  }
+
+  UidRecord.prototype.getPossibleRetryDelay = function(action) {
+
```
vladikoff: remove line break

lib/uid_record.js
```diff
@@ -0,0 +1,95 @@
...
+    if(this.timestampsByAction[action].length < limits.maxChecksPerUid){
+      this.rateLimitedTimestamp = undefined
+      return 0
+    } else {
+      if(!this.isRateLimited()){
+        //we have more attempts than allowed. rateLimit it
```
vladikoff: add space after `//`

test/remote/too_many_authenticated_checks.js
```diff
@@ -0,0 +1,166 @@
+/* Any copyright is dedicated to the Public Domain.
+ * http://creativecommons.org/publicdomain/zero/1.0/ */
+
+// Override limit values for testing
+
```
vladikoff: remove line break

test/remote/too_many_authenticated_checks.js
```diff
@@ -0,0 +1,166 @@
...
+        t.end()
+      }
+    )
+  }
+)
+
+test(
+  '/checkAuthenticated with same action',
+  function (t) {
+
```
vladikoff: remove line break

test/remote/too_many_authenticated_checks.js
```diff
@@ -0,0 +1,166 @@
...
+        t.fail(err)
+        t.end()
+      })
+  }
+)
+
+test(
+  '/checkAuthenticated with different actions',
+  function (t) {
+
```
vladikoff: remove line break

test/remote/too_many_authenticated_checks.js
```diff
@@ -0,0 +1,166 @@
...
+        return client.postAsync('/checkAuthenticated', { action: ACTION_B, ip: TEST_IP, uid: TEST_UID})
+      })
+      .spread(function(req, res, obj){
+        t.equal(res.statusCode, 200, 'returns a 200')
+        t.equal(obj.block, false, 'not rate limited on check1, actionB')
+
+        return client.postAsync('/checkAuthenticated', { action: ACTION_A, ip: TEST_IP, uid: TEST_UID})
+      })
+
```
vladikoff: remove line break

test/remote/too_many_authenticated_checks.js
```diff
@@ -0,0 +1,166 @@
...
+        return client.postAsync('/checkAuthenticated', { action: ACTION_A, ip: TEST_IP, uid: TEST_UID})
+      })
+      // uid should be now blocked to action1
+      .spread(function(req, res, obj){
+        t.equal(res.statusCode, 200, 'returns a 200')
+        t.equal(obj.block, true, 'uid is actionA rate limited after check3')
+
+        return client.postAsync('/checkAuthenticated', { action: ACTION_B, ip: TEST_IP, uid: TEST_UID})
+      })
+
```
vladikoff: remove line break

test/remote/too_many_authenticated_checks.js
```diff
@@ -0,0 +1,166 @@
...
+      .spread(function(req, res, obj){
+        t.equal(res.statusCode, 200, 'returns a 200')
+        t.equal(obj.block, false, 'not rate limited for actionB after check2')
+
+        // Delay ~2s for rate limit to go away
+        return Promise.delay(2010)
+      })
+
```
vladikoff: remove line break

vladikoff:
@rfk any feedback here?

lib/uid_record.js
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * This module keeps track of events related to a specific uid
+ */
+module.exports = function (limits, now) {
+
+  now = now || Date.now
+  function UidRecord() {
+    this.timestampsByAction = {}
+    this.rateLimitedTimestamp = undefined
```
rfk: Even though it's going to make the stored JSON bigger, this is so much better with readable field names rather than abbreviations like `this.bk` and `this.rl`; very nice :-)

lib/uid_record.js
```diff
@@ -0,0 +1,95 @@
...
+    this.trimCheckAuthenticated(action, timestamp)
+
+    if(!this.isRateLimited()){
+      this.timestampsByAction[action].push(timestamp)
+    }
+
+    return this.getPossibleRetryDelay(action)
+  }
+
+  UidRecord.prototype.trimCheckAuthenticated = function (action, now) {
```
rfk: nit re: the naming of this function, it may be easier to follow if named more directly for what it does, such as `trimCounts` by analogy to `addCount`.

test/memcache-helper.js
```diff
@@ -19,7 +19,14 @@ var config = {
     maxBadLogins: 2,
     maxBadLoginsPerIp: Number(process.env.MAX_BAD_LOGINS_PER_IP) || 3,
     ipRateLimitIntervalSeconds: Number(process.env.IP_RATE_LIMIT_INTERVAL_SECONDS) || 60 * 15,
-    ipRateLimitBanDurationSeconds: Number(process.env.IP_RATE_LIMIT_BAN_DURATION_SECONDS) || 60 * 15
+    ipRateLimitBanDurationSeconds: Number(process.env.IP_RATE_LIMIT_BAN_DURATION_SECONDS) || 60 * 15,
+    badLoginLockout: 3,
+    badLoginLockoutIntervalSeconds: 20,
```
rfk: These lockout-related settings have been removed in master, I wonder if they came back in through a rebase or something.  Anyway, they're not needed, please remove.

test/remote/too_many_login_checks.js
```diff
@@ -91,7 +91,7 @@ test(
         t.equal(res.statusCode, 200, 'returns a 200')
         t.equal(obj.block, true, 'ip is still rate limited')
 
-        // Delay ~3s for rate limit to go away
+        // Delay ~5s for rate limit to go away
```
rfk: nice catch :-)

lib/config/config.js
rfk: Seems nice and generous for a default value.

I wonder if we'll need different limits for different types of action...but I guess we can always do that in a follow-up if/when we need it.

test/memcache-helper.js
```diff
@@ -19,7 +19,14 @@ var config = {
...
+    uidRateLimit: {
+      limitIntervalSeconds: Number(process.env.UID_RATE_LIMIT_INTERVAL_SECONDS) || 60 * 15,
+      banDurationSeconds: Number(process.env.UID_RATE_LIMIT_BAN_DURATION_SECONDS) || 60 * 15,
+      maxChecks: Number(process.env.UID_RATE_LIMIT) || 3
+    }
```
rfk: I have a slight quibble about the inconsistency of using a nested `uidRateLimit ` object here, whereas e.g. the very similar `ipRateLimit` settings are all in flat top-level properties.  But just noting it, I don't feel strongly enough about it to block on it.

rfk:
I left a couple of nits, but this LGTM!

@vladikoff @larissagaulia I notice that we don't really use the IP for anything in `/checkAuthenticated`, do you think it's worth filing a follow-up issue to e.g. rate-limit a single IP that's doing actions against lots of authenticated accounts?

test/memcache-helper.js
vladikoff: > are all in flat top-level properties

Yeah it's different but I'm hoping we can nest future once like we do `uidRateLimit`

vladikoff:
Good work @larissagaulia 👍

PR #122, opened 2016-07-27
feat(server): Remove account lockout.

### LINKED ISSUE #120: Remove the account lockout feature
With signin confirmation and email caphta we have more targeted
meant to keep bad people at bay while minimizing friction for valid
users, account lockout no longer seems particularly elegant.

Ref https://github.com/mozilla/fxa-content-server/issues/3949
Ref https://github.com/mozilla/fxa-auth-server/issues/1359

shane-tomlinson:
fixes #120 

I made this into two commits. The first commit removes all references to `isWayOverBadLogins` and anywhere where `lockout` is referenced. The second commit also removes `lf` (loginFailures) from EmailRecord. I was less sure whether this would be used in the future for some feature, perhaps email captcha. I suppose we can always go back to this commit and update as needed.

@rfk and @seanmonstar - r?

### DIFF
--- a/lib/email_record.js
+++ b/lib/email_record.js
@@ -38,11 +38,6 @@ module.exports = function (limits, now) {
     return this.xs.length > limits.maxEmails
   }
 
-  EmailRecord.prototype.isWayOverBadLogins = function () {
-    this.trimBadLogins(now())
-    return this.lf.length > limits.badLoginLockout
-  }
-
   EmailRecord.prototype.trimHits = function (now) {
     if (this.xs.length === 0) { return }
     // xs is naturally ordered from oldest to newest
--- a/lib/email_record.js
+++ b/lib/email_record.js
@@ -11,7 +11,6 @@ module.exports = function (limits, now) {
 
   function EmailRecord() {
     this.xs = []
-    this.lf = []
   }
 
   EmailRecord.parse = function (object) {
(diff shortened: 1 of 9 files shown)


lib/email_record.js
```diff
@@ -20,16 +19,14 @@ module.exports = function (limits, now) {
     rec.bk = object.bk       // timestamp when the account was banned
     rec.rl = object.rl       // timestamp when the account was rate-limited
     rec.xs = object.xs || [] // timestamps when emails were sent
-    rec.lf = object.lf || [] // timestamps when a login failure occurred
```
shane-tomlinson: I was unsure whether we would want to keep `lf` around for email captcha, but figured we could re-add it with minimal fuss.

test/remote/too_many_bad_logins.js
```diff
@@ -119,166 +114,6 @@ test(
...
-                                                                                      function (err, req, res, obj) {
-                                                                                        t.equal(res.statusCode, 200, 'failed login 21')
-                                                                                        t.equal(obj.lockout, true, 'locked out')
-
-                                                                                        mcHelper.badLoginCheck(
-                                                                                          function (isOverBadLogins, isWayOverBadLogins) {
-                                                                                            t.equal(isOverBadLogins, false, 'is still not over bad logins')
-                                                                                            t.equal(isWayOverBadLogins, true, 'is now locked out')
-                                                                                            t.end()
```
rfk: I'm excited to see the removal of this Giant Indentation Pyramid Of Doom

seanmonstar: omg what is this. this is gold

seanmonstar:
\o/

rfk:
> The second commit also removes lf (loginFailures) from EmailRecord

I am surprised that we were *only* using this for the lockout stuff.  I guess we have the per-ip rate-limiting on logins, but I wonder if we should introduce a "no more than X logins in Y seconds on any one email address" rule that's independent of the IP.  @shane-tomlinson thoughts?

(Obviously that can be done in a separate issue, since we don't have such a rule right now anyway - it's just that this diff made me notice it)

PR #123, opened 2016-07-29
refactor(test): Modify test cases to use promises instead of callbacks

### LINKED ISSUE #97: Modify test cases to use promises instead of callbacks
Using promises could help clean up code and make it more readable.

https://github.com/mozilla/fxa-customs-server/blob/master/test/remote/too_many_bad_logins.js#L123

larissagaulia:
fixes #97


vladikoff:
Thanks!! 👍

PR #124, opened 2016-08-02
chore(server): Remove some left-over references to account lockout.

rfk:
A couple of stray references to account lockout that I noticed in passing; @shane-tomlinson r?


lib/actions.js
```diff
@@ -29,7 +27,6 @@ var ACCOUNT_STATUS_ACTION = {
 // us look like spammers if abused.
 var EMAIL_SENDING_ACTION = {
   accountCreate: true,
-  accountUnlockResendCode: true,
```
vladikoff: 👍

PR #129, opened 2016-09-07
feature(newrelic): add optional newrelic integration

jrgm:
Now for customs-server...

svcops would like to make use of New Relic. This PR just makes it possible to load the module if explicitly configured. Further configuration will happen with puppet in the circusd environment (unless we have to set so many switches that a sane configuration file makes sense).

r? @rfk


rfk:
LGTM, r+

PR #131, opened 2016-09-25
feat(unblock): add unblock rate limits

seanmonstar:
cc https://github.com/mozilla/fxa-auth-server/pull/1471

### DIFF
--- a/lib/email_record.js
+++ b/lib/email_record.js
@@ -12,6 +12,7 @@ module.exports = function (limits, now) {
   function EmailRecord() {
     this.vc = []
     this.xs = []
+    this.ub = []
   }
 
   EmailRecord.parse = function (object) {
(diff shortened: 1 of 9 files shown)


lib/email_record.js
```diff
@@ -11,15 +11,17 @@ module.exports = function (limits, now) {
...
+    this.ub = []
   }
 
   EmailRecord.parse = function (object) {
     var rec = new EmailRecord()
     object = object || {}
     rec.bk = object.bk       // timestamp when the account was banned
     rec.rl = object.rl       // timestamp when the account was rate-limited
-    rec.xs = object.xs || [] // timestamps when emails were sent
+    rec.xs = object.xs || rec.xs // timestamps when emails were sent
```
rfk: Does this fix a bug, or just for consistency?

rfk:
@seanmonstar as an unsolicited f?, this LGTM at a high-level - it's more ad-hoc coupling between auth-server and customs-server APIs, but we're well down that road already, so 👍

lib/email_record.js
seanmonstar: I just noticed that an Array is already initialized in the constructor, why create a new one. One might hope that V8 can initialize arrays for nothing, but :shrug:

rfk: 👍 ; mostly just confirming it wasn't a bug :-)

lib/server.js
```diff
@@ -118,15 +118,16 @@ module.exports = function createServer(config, log) {
...
               return {
                 block: true,
                 retryAfter: ipRecord.retryAfter()
               }
             }
 
 
-            var blockEmail = emailRecord.update(action)
+            var wantsUnblock = req.body.payload && req.body.payload.unblockCode
+            var blockEmail = emailRecord.update(action, !!wantsUnblock)
```
rfk: I think it would be worth adding a remote test exercises this payload-checking logic

seanmonstar:
@rfk added a remote test that `payload.unblockCode` affects the `EmailRecord` in memcached.

PR #133, opened 2016-09-29
fix(settings): Fix reloading of nested settings from mecmached

rfk:
Previously this was not pushing the `uidRateLimit` settings object into memcached, so it was failing when it tried to load it back out.  @vbudhram r?

### DIFF
--- a/test/remote/config_update_tests.js
+++ b/test/remote/config_update_tests.js
@@ -56,10 +56,11 @@ test(
       })
       .then(function (bis) {
         x = bis
-        return mcHelper.setLimits({ blockIntervalSeconds: bis + 1, uidRateLimit:{} })
+        return mcHelper.setLimits({ blockIntervalSeconds: bis + 1 })
       })
       .then(function (settings) {
         t.equal(x + 1, settings.blockIntervalSeconds, 'helper sees the change')
+        // Wait for background polling to detect the new value in memcache
         return Promise.delay(1010)
       })
       .then(function() {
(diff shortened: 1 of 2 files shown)


test/remote/config_update_tests.js
```diff
@@ -56,7 +56,7 @@ test(
       })
       .then(function (bis) {
         x = bis
-        return mcHelper.setLimits({ blockIntervalSeconds: bis + 1, uidRateLimit:{} })
+        return mcHelper.setLimits({ blockIntervalSeconds: bis + 1 })
```
rfk: This appears to have been a hack around this problem, it can be removed now that the nested object is handled correctly.

rfk:
The decrease in code-coverage here, is because it's no longer exercising the various error-logging code paths in limits.js.  Previously it was logging an error due to non-fatal failures to update the limits config in the background.

test/remote/config_update_tests.js
```diff
@@ -77,6 +77,37 @@ test(
...
+        // This is derived from uidRateLimit.maxChecks
+        return obj.maxChecksPerUid
+      })
+      .then(function (mcpuid) {
+        x = mcpuid
+        return mcHelper.setLimits({ uidRateLimit: { maxChecks: mcpuid + 1 } })
+      })
+      .then(function (settings) {
+        t.equal(x + 1, settings.maxChecksPerUid, 'helper sees the change')
+        return Promise.delay(1010)
```
vbudhram: Do you know why we have delays here? Removing them causes tests to fail and I couldn't find any comments on why they are needed.

rfk: It's to wait for the server's background polling to detect the new value in memcache, and update itself accordingly.  I'll add a comment :-)

Issue #126, opened 2016-08-10
Backport code refactor from "request checks" private fork


rfk:
The "request checks" work on the private repo included a refactor of the flow of control in `lib/server.js`:

  https://github.com/mozilla/fxa-customs-server-private/commit/12ea73dca58df8dbd6f43b7d43189c9a9746f4ba

We should backport it to the public repo so that it's easier to manage patches between the two (but without copying across the actual checks, which have to remain secret).

PR #137, opened 2016-10-14
feat(requestChecks): Backport "requestChecks" framework from private repo

rfk:
Fixes https://github.com/mozilla/fxa-customs-server/issues/126

@vbudhram r?

### DIFF
--- a/lib/requestChecks.js
+++ b/lib/requestChecks.js
@@ -1,0 +1,81 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var deepEqual = require('deep-equal')
+
+module.exports = function (config, mc, log) {
+
+  var requestChecks = null
+  var pollInterval = null
+
+  function RequestChecks(settings) {
+    this.setAll(settings)
+  }
+
+  RequestChecks.prototype.setAll = function (settings) {
+    this.treatEveryoneWithSuspicion = settings.treatEveryoneWithSuspicion
+    // The private branch puts some additional private config here.
+    return this
+  }
+
+  RequestChecks.prototype.push = function () {
+    log.info({ op: 'requestChecks.push' })
+    return mc.setAsync('requestChecks', this, 0)
+      .then(this.refresh.bind(this))
+  }
+
+  RequestChecks.prototype.refresh = function (options) {
+    log.info({ op: 'requestChecks.refresh' })
+    var result = mc.getAsync('requestChecks').then(validateAndMerge)
+
+    if (options && options.pushOnMissing) {
+      result = result.catch(this.push.bind(this))
+    }
+
+    return result.then(
+      this.setAll.bind(this),
+      function (err) {
+        log.error({ op: 'requestChecks.refresh', err: err })
+      }
+    )
+  }
+
+  RequestChecks.prototype.pollForUpdates = function () {
+    this.stopPolling()
+    pollInterval = setInterval(this.refresh.bind(this), config.updatePollIntervalSeconds * 1000)
+  }
+
+  RequestChecks.prototype.stopPolling = function () {
+    clearInterval(pollInterval)
+  }
+
+  // Type-checks updates to the settings, and merges them
+  // with the current values, modying its argument in-place.
+  function validateAndMerge(settings) {
+    if (typeof(settings) !== 'object') {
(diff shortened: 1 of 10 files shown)


rfk:
@vbudhram this is the backport work we talked about earlier this week.  I just went ahead and did it because the train-71 merge to private repo as busted, and this will help prevent bustage in the future.

lib/server.js
```diff
@@ -145,51 +150,68 @@ module.exports = function createServer(config, log) {
...
             // IP's that are in blocklist should be blocked
             // and not return a retryAfter because it is not known
             // when they would be removed from blocklist
-            if (config.ipBlocklist.enable && blockListManager.contains(ip) && !allowedIP && !allowedEmailDomain) {
+            if (config.ipBlocklist.enable && blockListManager.contains(ip)) {
               if (!config.ipBlocklist.logOnly) {
-
-                // When blocklist is enabled, explicitly skip setting records.
-                // Requests issued while on blocklist will not count towards
-                // the rate limit.
```
rfk: I'm not sure there's too much value in avoiding the write here, and removing the `return` means that there's now a single place where we return from this function, and can check that we include all the relevant fields in the response.  I'm open to putting it back in though.

lib/server.js
```diff
@@ -145,51 +150,68 @@ module.exports = function createServer(config, log) {
...
         )
         .then(
           function (result) {
-            log.info({ op: 'request.check', email: email, ip: ip, action: action, block: result.block })
+            // Regardless of all the above checks, there are some
+            // IPs and emails that we just won't block.  These are
+            // typically for Mozilla QA purposes.  We check them at
+            // the end so as not to pay the overhead of the checks
+            // on the many requests that are *not* QA-related.
+            if (result.block || result.suspect) {
```
vbudhram: 👍

lib/requestChecks.js
```diff
@@ -0,0 +1,79 @@
...
+  RequestChecks.prototype.pollForUpdates = function () {
+    this.stopPolling()
+    pollInterval = setInterval(this.refresh.bind(this), config.updatePollIntervalSeconds * 1000)
+  }
+
+  RequestChecks.prototype.stopPolling = function () {
+    clearInterval(pollInterval)
+  }
+
+  function validate(settings) {
```
vbudhram: nit: This function seems to do a little more than just validate. Possibly add comment to mention that it returns updated settings based on `config.requestChecks`?

lib/server.js
vbudhram: I was on the fence about which this should go originally, I'm cool with it being handled in a single spot.

lib/server.js
```diff
@@ -145,51 +150,68 @@ module.exports = function createServer(config, log) {
...
+                result.block = false
+                result.suspect = false
+              }
+              if (allowedEmailDomains.isAllowed(email)) {
+                result.block = false
+                result.suspect = false
+              }
+            }
+
+            log.info({
```
vbudhram: Is there any value in us logging that a request would have been blocked but it wasn't because it was whitelisted?

lib/server.js
```diff
@@ -145,51 +150,68 @@ module.exports = function createServer(config, log) {
...
           }
         )
         .then(
           function (result) {
-            log.info({ op: 'request.check', email: email, ip: ip, action: action, block: result.block })
+            // Regardless of all the above checks, there are some
+            // IPs and emails that we just won't block.  These are
+            // typically for Mozilla QA purposes.  We check them at
+            // the end so as not to pay the overhead of the checks
+            // on the many requests that are *not* QA-related.
```
rfk: The checking of `allowedIPs` and `allowedEmailDomains` has been pulled out into a separate function, so that we can be sure it's always executed regardless of what happens in the logic above.

scripts/config-list.js
```diff
@@ -58,6 +60,24 @@ mc.getAsync('limits')
       else {
         output.allowedIPs = data
       }
+      return mc.getAsync('allowedEmailDomains')
+    }
+  )
+  .then(
+    function (data) {
+      if (!data) { console.error('no allowedEmailDomains set') }
```
vbudhram: Haven't seen if statements like this in any other project, but it is consistent with this file :)

lib/server.js
rfk: Yes, I like it it, will add.

scripts/config-list.js
rfk: @vbudhram I pushed a second commit that fixes this to be consistent with our usual style.

lib/requestChecks.js
rfk: I renamed it and added a bit of commentary

PR #138, opened 2016-10-17
chore(config): Don't set allow ALLOWED_IPS by default.

rfk:
Per conversation in https://github.com/mozilla-services/puppet-config/issues/2330, let's default this to an empty list to make life easier for Ops.  They were overriding these values in production anyway.


rfk:
@jrgm r?

jrgm:
r+

I'll need to remove settings in puppet-config too.

PR #141, opened 2016-10-18
fix(logging): Don't attempt to log a 'msg' field.

### LINKED ISSUE #140: fix logging of `msg` member
In production logs, I get `"msg":"",`. However, I know that `msg` is not the empty string in the code. I don't know where this is getting lost, but it should be in the logs.

rfk:
The logger used here (bunyan) treats 'msg' as a place to write its own formatted interpretation of the log message, so don't use it for our own purposes.  Instead let's log it under the name "message".  Bleh, I know, but simplest fix to get production logging up and running in the next train.  @jrgm r?

Fixes #140

### DIFF
--- a/lib/limits.js
+++ b/lib/limits.js
@@ -79,7 +79,7 @@ module.exports = function (config, mc, log) {
       var current = limits[key]
       var future = settings[key]
       if (typeof(current) !== typeof(future)) {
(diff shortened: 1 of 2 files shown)


rfk:
(I tested this locally by putting some bad config values in memcached, and ensuring that the error logs included the expected message)

jrgm:
Nice sleuthing. LGTM!

PR #143, opened 2016-10-18
fix(blocklist): Convert date to milliseconds for file comparison

### LINKED ISSUE #142: File mtime is a Date, not a Number
Noticed in the code - https://github.com/mozilla/fxa-customs-server/blob/master/lib/ip_blocklist.js#L185

`mtime` on `fs.Stats` is a `Date` not a `Number`. So that comparison works, but probably a `toString` comparison is not what was intended given https://github.com/mozilla/fxa-customs-server/blob/master/lib/ip_blocklist.js#L52

vbudhram:
Per https://nodejs.org/api/fs.html#fs_class_fs_stats, uses `getTime()` on mtime, for use in comparisons.

Fixes #142

### DIFF
--- a/lib/ip_blocklist.js
+++ b/lib/ip_blocklist.js
@@ -67,7 +67,7 @@ module.exports = function (log, config) {
     var startTime = Date.now()
     return statFile(filePath)
       .then(function (fileStats) {
-        self.fileLastModified = fileStats.mtime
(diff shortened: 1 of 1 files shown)


vbudhram:
@jrgm Mind an r?

rfk:
LGTM!  Waiting for travis to re-build for sporadic test failure.

PR #145, opened 2016-10-25
fix(ip_record): Correctly total bad logins by unique email address.

rfk:
As noted in https://github.com/mozilla/fxa-content-server/issues/4324, the per-ip-address login rate-limiting is over-counting failed logins from a single email address.  This fixes it to only take the max-weighted errno from each unique email address.

Amusingly, this bug was causing the existing "addBadLogins works per IP" to pass, when it should have been failing.

@shane-tomlinson since this unblocks signin unblock, r?

### DIFF
--- a/lib/ip_record.js
+++ b/lib/ip_record.js
@@ -39,17 +39,21 @@ module.exports = function (limits, now) {
   IpRecord.prototype.isOverBadLogins = function () {
     this.trimBadLogins(now())
     // IPs are limited based on the number of unique email
-    // addresses they access.  Take the highest-weighted
-    // bad-login event for each email address.
-    var total = 0
-    var seen = {}
+    // addresses they access.  Sum the highest-weighted
+    // bad-login event for each user account to determine
+    // the overall bad-logins score.
+    var weights = {}
     this.lf.forEach(function(info) {
-      var incr = limits.badLoginErrnoWeights[info.e] || 1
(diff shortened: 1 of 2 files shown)


lib/ip_record.js
```diff
@@ -41,15 +41,16 @@ module.exports = function (limits, now) {
...
-    var total = 0
-    var seen = {}
+    var weights = {}
     this.lf.forEach(function(info) {
-      var incr = limits.badLoginErrnoWeights[info.e] || 1
-      if (info.u in seen && seen[info.u] < incr) {
-        total -= seen[info.u]
-        seen[info.u] = incr
-      }
-      total += incr
```
rfk: The previous code here was trying to be smart and keep a running total by subtracting the old weight and adding the new one, but it's buggy.  So I replaced it with two separate loops that more obviously do the right thing.

shane-tomlinson: @rfk - can you add in comments the intent of the code and maybe a short description of the algorithm? The ultra-terse variable names leave me scratching my head, I have no idea what `e` or `u` are.

rfk:
@seanmonstar mind a quick r? of this to unblock train-72?

rfk:
Thanks @seanmonstar; trying to figure out the travis bustage here before merge

seanmonstar:
Looks like travis is happy now!

PR #146, opened 2016-11-17
Highly experimental re-implementation of ip blocklist datastructure

rfk:
My internet was down for like 4 hours today, so I took the opportunity to dust off a little side-project I've been working on whenever I need to clear my head.  It's a highly experimental re-write of the data-structure used to manage our IP blocklists, to try to make it faster and more memory-friendly.

I'm not really suggesting we merge this any time soon, as it's pretty complex and not well tested.  But I thought I'd put it up out of interest. 

Here's the fun part, some benchmarks.  First, the existing blocklist code:

```
$ node --expose-gc ./scripts/benchmark_ip_blocklist.js ./lib/ip_blocklist.js ~/Desktop/firehol_level4.netset
Base memory usage of this script: 25600000
Memory usage after loading the blocklist: 73777152
Checked 10000 ips, found 31 in the blocklist
Best time of 5 runs: 1065 ms
Memory usage after running the test: 74207232
```

That's not bad, although it uses a fair bit of memory for the large level4 list.  Here's the new version, which uses a densely-packed integer array with a radix-tree index for its lookups:

```
$ node --expose-gc ./scripts/benchmark_ip_blocklist.js ./lib/ip_blocklist_critbit.js ~/Desktop/firehol_level4.netset
Base memory usage of this script: 25878528
Memory usage after loading the blocklist: 62017536
Checked 10000 ips, found 31 in the blocklist
Best time of 5 runs: 29 ms
Memory usage after running the test: 62255104
```

That's significantly faster \o/

It also uses a smaller amount of steady-state memory to hold the list.  Interestingly, the *peak* memory usage is very similar, since both versions pull the whole list into memory and do a bunch of preprocessing on it:

```
$ /usr/bin/time -l node --expose-gc ./scripts/benchmark_ip_blocklist.js ./lib/ip_blocklist.js ~/Desktop/firehol_level4.netset 2>&1 | grep "resident set size"
 105291776  maximum resident set size
```

Versus:

```
$ /usr/bin/time -l node --expose-gc ./scripts/benchmark_ip_blocklist.js ./lib/ip_blocklist_critbit.js ~/Desktop/firehol_level4.netset 2>&1 | grep "resident set size"
 105107456  maximum resident set size
```

But the new version allows most of that to be GC'd, leaving behind just single ArrayBuffer with the search data in it.

If you cache this pre-processed form of the blocklist on disk, we can just slurp it straight into an ArrayBuffer and the peak memory usage is significantly lower:

```
rfk@tangello:fxa-customs-server(ip-blocklist-critbit *)$ /usr/bin/time -l node --expose-gc ./scripts/benchmark_ip_blocklist.js ./lib/ip_blocklist_critbit.js ~/Desktop/firehol_level4.netset 2>&1 | grep "resident set size"
  32546816  maximum resident set size
```

Anyways, a fun experiment that I thought would be worth sharing.  And hey, if the blocklists turn out to be valuable but slow, it may be worth our while to try to make this production ready one day :-)

### DIFF
--- a/lib/ip_blocklist_critbit.js
+++ b/lib/ip_blocklist_critbit.js
@@ -1,0 +1,428 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * This module contains all the functionality needed to initialize, load
+ * and check an ip blocklist. The ip blocklist file must contain
+ * at least one column that has ip addresses or an ip address range
+ * in the following formats, 192.168.0.0 or 192.168.0.0/32. The blocklist
+ * polls for any file changes and reloads itself if needed.
+ *
+ */
+var Promise = require('bluebird')
+var readFile = Promise.promisify(require('fs').readFile)
+var writeFile = Promise.promisify(require('fs').writeFile)
+var statFile = Promise.promisify(require('fs').stat)
+var parse = Promise.promisify(require('csv-parse'))
+var path = require('path')
+var ip = require('ip')
+
+
+
+/**
+ * These functions allow for quickly checking if an IPv4 addresses
+ * is a member of a fixed set.  There's one that prepares the data
+ * for lookups, and one that makes a fast matching function from the
+ * prepared data:
+ *
+ *    var dataBuffer = makeIPv4MatcherData([...list IP strings...])
+ *    var check = makeIPv4MatcherFunc(dataBuffer)
+ *    if (check("1.2.3.4")) {
+ *      console.log("it was in the set!")
+ *    }
+ *
+ * They're split up like this so that you can do the preparation
+ * ahead of time, and also to allow the matching function to be
+ * as efficient as possible.
+ *
+ * We represent a set of IPv4 CIDRs as an array of 32-bit integers
+ * storing both the values themselvs, and an index through which to
+ * search on them.  The idea is to keep all of the data densely packed
+ * in memory, so while things will work using an ordinary javascript
+ * array, you  might like to use an ArrayBuffer instead.
+ *
+ * The structure of the data is [metadata][index][values], as
+ * follows:
+ *
+ *  - The first metadata item, data[0], gives the offset of the
+ *    end of the index, i.e. the location of the first value in
+ *    in the set.
+ *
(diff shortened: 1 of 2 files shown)


rfk:
Fun trivia: most of the speed increase here is from how quickly we can find IPs that are *not* in the blocklist.  If I change the test to only search for an IP that's in the blocklist, the times are closer:

```
$ node --expose-gc ./scripts/benchmark_ip_blocklist.js ./lib/ip_blocklist.js ~/Desktop/firehol_level4.netset
Base memory usage of this script: 25870336
Memory usage after loading the blocklist: 73732096
Checked 10000 ips, found 10000 in the blocklist
Best time of 5 runs: 235 ms
Memory usage after running the test: 73912320

```

Versus:

```
$ node --expose-gc ./scripts/benchmark_ip_blocklist.js ./lib/ip_blocklist_critbit.js ~/Desktop/firehol_level4.netset
Base memory usage of this script: 25759744
Memory usage after loading the blocklist: 61857792
Checked 10000 ips, found 10000 in the blocklist
Best time of 5 runs: 25 ms
Memory usage after running the test: 61935616
```

lib/ip_blocklist_critbit.js
```diff
@@ -0,0 +1,428 @@
...
+ *
+ *      - the bottom 5 bits give the number of bits to shift
+ *        the bitmask by for the next lookup (1 through 31).
+ *    
+ *      - the remaining 26 bits give the offset of the next
+ *        index node to look at, from the start of the data.
+ *
+ *  - When examining input bit b at an internal node
+ *    with offset n, we look at data[n] when input[b]
+ *    is zero and data[n+1] when input[b] is one.
```
rfk: Also this could probably benefit from a diagram or two...

jvehent:
This is cool! I'll have to double-check with @autrilla, but I'm pretty sure the ip reputation service uses a radix-tree-like index in postgres to achieve similar lookup performances.

autrilla:
Yeah, we use ip4r's indexes, which underneath use radix trees, and they're super fast. A lookup for a single IP was ~0.2ms in postgres, but I didn't try looking for matches in a list of IPs.

lib/ip_blocklist_critbit.js
vbudhram: I read this sectio a few times but still had trouble getting my head around it, a diagram would help greatly.

rfk:
This was a lot of fun, but I think it's pretty obvious we're not going to do anything with it, so closing :-)

PR #150, opened 2016-11-24
chore(shrinkwrap): add npm script for shrinkwrap

### LINKED ISSUE #149: Add an `npm run shrinkwrap` command
It should "do the right thing" so we can stop thinking about it.

Ref 
https://github.com/mozilla/fxa-content-server/issues/4439
https://github.com/mozilla/fxa-auth-server/issues/1564
https://github.com/mozilla/fxa-profile-server/issues/223
https://github.com/mozilla/fxa-auth-db-mysql/issues/189

Example: https://github.com/mozilla/fxa-auth-mailer/blob/master/package.json#L10

TDA:
This commit updates the npm script for generating shrinkwrap and moves the nsp into its own command.
Fixes #149 

Not sure why nsp was present in the first place, so moved it into its own command :)

@vladikoff @shane-tomlinson

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -2,41 +2,173 @@
   "name": "fxa-customs-server",
   "version": "0.72.1",
   "dependencies": {
-    "ansi-regex": {
-      "version": "2.0.0",
-      "from": "ansi-regex@>=2.0.0 <3.0.0",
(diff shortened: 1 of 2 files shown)


vladikoff:
see comment in other repo

vladikoff:
everything should use 'npmshrink' except content server, which adds :prod to the command

TDA:
@vladikoff This is updated too :)

PR #152, opened 2016-12-16
feat(ipreputation): Use IP reputation service from /check

g-k:
Add feature flag to check reputation IPs on the `/check` route and block requests below `REPUTATION_SERVICE_BLOCK_BELOW`.

Functional Tests:

setup:

* run memcached and postgres-ip4r docker container
* run tigerblood: `TIGERBLOOD_DSN="user=tigerblood dbname=tigerblood sslmode=disable" ./tigerblood`
* run customs server: `REPUTATION_SERVICE_ENABLE=true REPUTATION_SERVICE_ENABLE_CHECK=true REPUTATION_SERVICE_BLOCK_BELOW=50 node bin/customs_server.js`

Check `/check` against running tigerblood server:

```
# IP without reputation

curl -w '\n' -v -H "Content-Type: application/json" "http://127.0.0.1:7000/check" -d '{"email": "me@example.com", "ip
": "192.0.2.1", "action": "accountCreate"}'
{"block":true,"retryAfter":900,"unblock":true,"suspect":false}

## customs server logs

{"name":"customs-server","hostname":"gguthe-23818","pid":16752,"level":20,"op":"fetchIPReputation","ip":"192.0.2.1","err":"Reputation not found for IP.","msg":"","time":"2016-12-16T17:46:33.268Z","v":0}
{"name":"customs-server","hostname":"gguthe-23818","pid":16752,"level":30,"op":"request.check","email":"me@example.com","ip":"192.0.2.1","action":"accountCreate","block":true,"unblock":true,"suspect":false,"msg":"","time":"2016-12-16T17:46:33.272Z","v":0}

# IP with unblocked reputation

## created by previous request
curl -w '\n' -v -H "Content-Type: application/json" "http://127.0.0.1:8080/192.0.2.1"
{"IP":"192.0.2.1","Reputation":100}

## customs server logs

{"name":"customs-server","hostname":"gguthe-23818","pid":16752,"level":30,"op":"request.check","email":"me@example.com","ip":"192.0.2.1","action":"accountCreate","block":true,"unblock":true,"suspect":false,"msg":"","time":"2016-12-16T17:50:19.133Z","v":0}

# IP with blocked reputation

curl -X DELETE -w '\n' -v -H "Content-Type: application/json" "http://127.0.0.1:8080/192.0.2.1"
curl -w '\n' -v -H "Content-Type: application/json" "http://127.0.0.1:8080/" -d '{"ip": "192.0.2.1", "reputation": 9}
'

curl -w '\n' -v -H "Content-Type: application/json" "http://127.0.0.1:7000/check" -d '{"email": "me@example.com", "ip
": "192.0.2.1", "action": "accountCreate"}'
{"block":true,"retryAfter":0}

## customs server logs

{"name":"customs-server","hostname":"gguthe-23818","pid":16752,"level":30,"op":"request.check","email":"me@example.com","ip":"192.0.2.1","action":"accountCreate","block":true,"msg":"","time":"2016-12-16T17:51:37.680Z","v":0}
```

Check `/check` when tigerblood server isn't running (but check is enable):

```
# stop tigerblood

curl -w '\n' -v -H "Content-Type: application/json" "http://127.0.0.1:7000/c
heck" -d '{"email": "me@example.com", "ip": "192.0.2.1", "action": "accountCreate"}'
{"block":false,"retryAfter":0,"unblock":true,"suspect":false}

## customs server logs

{"name":"customs-server","hostname":"gguthe-23818","pid":16752,"level":50,"op":"fetchIPReputation","ip":"192.0.2.1","err":{"cause":{"code":"ECONNREFUSED","errno":"ECONNREFUSED","syscall":"connect","address":"127.0.0.1","port":8080},"isOperational":true,"code":"ECONNREFUSED","errno":"ECONNREFUSED","syscall":"connect","address":"127.0.0.1","port":8080},"msg":"","time":"2016-12-16T17:54:31.545Z","v":0}
{"name":"customs-server","hostname":"gguthe-23818","pid":16752,"level":30,"op":"request.check","email":"me@example.com","ip":"192.0.2.1","action":"accountCreate","block":false,"unblock":true,"suspect":false,"msg":"","time":"2016-12-16T17:54:31.548Z","v":0}
```

### DIFF
--- a/lib/config/config.js
+++ b/lib/config/config.js
@@ -219,6 +219,24 @@ module.exports = function (fs, path, url, convict) {
         default: false,
         env: 'REPUTATION_SERVICE_ENABLE'
       },
+      enableCheck: {
+        doc: 'Flag to enable fetching reputation for IPs on /check route',
+        format: Boolean,
+        default: false,
+        env: 'REPUTATION_SERVICE_ENABLE_CHECK'
+      },
+      suspectBelow: {
+        doc: 'Suspect /check requests from IPs with reputation less than this',
+        format: 'int',
+        default: 0,
+        env: 'REPUTATION_SERVICE_SUSPECT_BELOW'
+      },
+      blockBelow: {
+        doc: 'Block /check requests from IPs with reputation less than this',
+        format: 'int',
+        default: 0,
+        env: 'REPUTATION_SERVICE_BLOCK_BELOW'
+      },
       host: {
         doc: 'The reputation service IP address',
         default: '127.0.0.1',
--- a/lib/server.js
+++ b/lib/server.js
@@ -7,6 +7,7 @@
 var Memcached = require('memcached')
 var restify = require('restify')
 var packageJson = require('../package.json')
+var blockReasons = require('./block_reasons')
 var P = require('bluebird')
 P.promisifyAll(Memcached.prototype)
 
@@ -37,16 +38,7 @@ module.exports = function createServer(config, log) {
     }
   )
 
-  if (config.reputationService.enable) {
-    var IPReputationClient = require('ip-reputation-js-client')
-    var ipClient = new IPReputationClient({
-      host: config.reputationService.host,
-      port: config.reputationService.port,
-      id: config.reputationService.hawkId,
-      key: config.reputationService.hawkKey,
-      timeout: config.reputationService.timeout
-    })
-  }
+  var reputationService = require('./reputationService')(config, log)
   var limits = require('./limits')(config, mc, log)
   var allowedIPs = require('./allowed_ips')(config, mc, log)
   var allowedEmailDomains = require('./allowed_email_domains')(config, mc, log)
@@ -86,14 +78,14 @@ module.exports = function createServer(config, log) {
   }
 
   function fetchRecords(email, ip) {
-    return P.all(
-      [
-        // get records and ignore errors by returning a new record
-        mc.getAsync(email).then(EmailRecord.parse, EmailRecord.parse),
-        mc.getAsync(ip).then(IpRecord.parse, IpRecord.parse),
-        mc.getAsync(ip + email).then(IpEmailRecord.parse, IpEmailRecord.parse)
-      ]
-    )
+    var promises = [
+      // get records and ignore errors by returning a new record
+      mc.getAsync(email).then(EmailRecord.parse, EmailRecord.parse),
+      mc.getAsync(ip).then(IpRecord.parse, IpRecord.parse),
+      mc.getAsync(ip + email).then(IpEmailRecord.parse, IpEmailRecord.parse),
+      reputationService.get(ip)
+    ]
+    return P.all(promises)
   }
 
   function setRecord(key, record) {
@@ -138,7 +130,7 @@ module.exports = function createServer(config, log) {
 
       fetchRecords(email, ip)
         .spread(
-          function (emailRecord, ipRecord, ipEmailRecord) {
+          function (emailRecord, ipRecord, ipEmailRecord, reputation) {
             if (ipRecord.isBlocked()) {
               // a blocked ip should just be ignored completely
               // it's malicious, it shouldn't penalize emails or allow
@@ -162,8 +154,13 @@ module.exports = function createServer(config, log) {
             var retryAfter = [blockEmail, blockIpEmail, blockIp].reduce(max)
             var block = retryAfter > 0
             var suspect = false
+            var blockReason = null
 
-            if (requestChecks.treatEveryoneWithSuspicion) {
+            if (block) {
+              blockReason = blockReasons.OTHER
+            }
+
+            if (requestChecks.treatEveryoneWithSuspicion || reputationService.isSuspectBelow(reputation)) {
               suspect = true
             }
             // The private branch puts some additional request checks here.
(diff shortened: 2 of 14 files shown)


rfk:
Thanks @g-k!  I haven't had a thorough look at this yet, but a couple of high-level thoughts while I think them:

* We have the ability to "suspect" as well as "block", and it's probably worth using that here.  We could have two threshold, one for IPs that we outright block, and one for IPs that we'll allow through but treat with a bit of extra suspicion.

* Since we're both consuming and producing for this reputation service, we need to be careful about creating feedback loops here.  If we block you because you've got a bad reputation, does that get fed back to tigerblood and give you an even worse reputation?  Will that cause you to never be able to escape its judgement?  Is that in fact the desired effect?

g-k:
Great points!  Made some changes:

* a configurable reputation threshold to mark `/check` requests as suspect
* a blockReasons enum to not send a violation if the `/check` request was blocked due to insufficient reputation (generally I think we should avoid fan-out and broadcast storm type behavior between the services since we're trying to prevent DOS)
* fixed the send violations tests flickering
* ran remote tests in randomized file order a few times to see if they're depending on weird side effects `node scripts/tap-coverage.js $(ls -1f test/remote/*.js | gsort -R | tr '\n' ' ')`

Should we send unblock events to help restore reputation faster?

I have no idea what the perf impact this stuff might have and no data that multiple services might be attacked from the same IPs.

rfk:
> Should we send unblock events to help restore reputation faster?

I think this would be valuable, although IIRC customs-server can currently only see if there's an unblock *attempt*, not necessary a successful unblock.

g-k:
Sounds good.  Tigerblood doesn't support positive reputation events.

r? @vbudhram or feel free to reassign.

lib/config/config.js
```diff
@@ -219,6 +219,24 @@ module.exports = function (fs, path, url, convict) {
         default: false,
         env: 'REPUTATION_SERVICE_ENABLE'
       },
+      enable_check: {
```
vladikoff: style: these new props need to be camelCase, like other config props in this file.

lib/server.js
```diff
@@ -85,15 +86,31 @@ module.exports = function createServer(config, log) {
...
-    )
+  function fetchRecords(email, ip, includeReputation) {
+    var promises = [
+      // get records and ignore errors by returning a new record
+      mc.getAsync(email).then(EmailRecord.parse, EmailRecord.parse),
+      mc.getAsync(ip).then(IpRecord.parse, IpRecord.parse),
+      mc.getAsync(ip + email).then(IpEmailRecord.parse, IpEmailRecord.parse)
+    ]
+    if (includeReputation) {
+      var fetchReputation = ipClient.get(ip)
```
vladikoff: nit: it might be beneficial to extract this new`fetchReputation` logic into its own function.

vbudhram: Along those lines, I think it might be useful to put the reputation logical into its own class. This class could handle all of the enable checks and reporting to the reputation service. This way we could keep the `server.js` file a little cleaner.

lib/server.js
```diff
@@ -162,8 +179,14 @@ module.exports = function createServer(config, log) {
...
             var block = retryAfter > 0
             var suspect = false
+            var blockReason = null
+
+            if (block) {
+              blockReason = blockReasons.OTHER
+            }
 
-            if (requestChecks.treatEveryoneWithSuspicion) {
+            if (requestChecks.treatEveryoneWithSuspicion
```
vbudhram: Similarly you could put the reputation check part of this into the class. Something like `reputationService.isSuspectBelow(reputation)`

lib/server.js
```diff
@@ -224,9 +256,16 @@ module.exports = function createServer(config, log) {
...
-            res.send(result)
+            res.send({
+              block: result.block,
+              retryAfter: result.retryAfter,
+              unblock: result.unblock,
+              suspect: result.suspect
+            })
 
-            if (config.reputationService.enable && result.block) {
+            if (config.reputationService.enable
```
vbudhram: This could be added to class as well, maybe as `reputationService.report(ip, result)`

test/remote/check_reputation_tests.js
```diff
@@ -0,0 +1,254 @@
...
+        t.end()
+      }).catch(function(err) {
+        t.fail(err)
+        t.end()
+      })
+  }
+)
+
+test(
+  'does not block /check for whitelisted IP with reputation below block_below',
```
vbudhram: 👍

test/remote/failed_login_attempt_tests.js
```diff
@@ -26,6 +27,18 @@ test(
   }
 )
 
+test(
+  'clear everything',
```
vbudhram: Thanks for adding these!

vbudhram:
@g-k LGTM, merging with feature disabled by default.

vbudhram:
@g-k Spoke a little to soon, noticed that the commit message does not follow the format of the other commits in this repo. Would you mind squashing and renaming the commit to `feat(ipreputation): Use IP reputation service` or something similar?

vbudhram:
@g-k Ran this several times both locally and in Travis with no test errors, looks resolved for now. We can reopen #153 if this occurs again.

PR #156, opened 2017-01-10
fix(config): remove mockmyid rate limit, add second restmail

vladikoff:
@rfk r?


rfk:
👍 apart form test failures :-)

PR #159, opened 2017-01-24
fix(retryAfter): Round blocking periods up instead of down.

### LINKED ISSUE #158: intermittent test failures in retryAfter
refs: https://github.com/mozilla/fxa-customs-server/issues/153 https://github.com/mozilla/fxa-customs-server/pull/152#pullrequestreview-16365892

[This branch](https://github.com/mozilla/fxa-customs-server/compare/master...g-k:flaky-email-record-test?expand=1) adds a test case that fails intermittently (for me at least).  In bash or zsh it can be run until failure (usually within ten seconds) with: `while node_modules/.bin/tap test/local/flaky_email_record_test.js; do :; done`

The test fails on v0.78.0 so the issue isn't limited to changes in https://github.com/mozilla/fxa-customs-server/pull/152.

Should apply to all rate limit tests hitting the `/check` route where `limits.rateLimitIntervalMs` is 1000 and now is `Date.now`, but I've only looked at `too_many_verify_codes`.

The failures seem to stem from computing `Math.floor(0.999)` when calculating [rateLimitAfter](https://github.com/mozilla/fxa-customs-server/blob/master/lib/email_record.js#L136).

Possible fixes:

* don't worry about it and rerun tests
* include a tolerance param in the comparision
* run flaky tests with acceptance criteria like 2/3 runs pass, so adding new rate limit tests don't increase the overall failure rate

#### Background

Working on https://github.com/mozilla/fxa-customs-server/pull/152, it was weird that  [too_many_verify_codes](https://travis-ci.org/mozilla/fxa-customs-server/jobs/191366059) and [too_many_account_status_check](https://travis-ci.org/mozilla/fxa-customs-server/jobs/191363227) failed in addition to tests I added (especially after [wiping memcache](https://github.com/mozilla/fxa-customs-server/commit/4f5d781e6dceb765ab96b8c8b305bb7023a55536#diff-29937e99d30174d1de102c03acaa8a22)).

The fix for https://github.com/mozilla/fxa-customs-server/issues/153 reduced the error rate for the send violations tests, but it still failed occasionally.

rfk:
Fixes #158; @vbudhram r?; /cc @g-k 

Previously, the code would round your ban period down to the
nearest integral number of seconds.  Since the ban duration in
tests is only 1 second, this would occasionally cause it to be
rounded down to zero, leading to flaky test failures.

We now round the ban priod *up* to the nearest integral number
of seconds, which shouldn't have any significant impact on users
in production, but makes the tests much simpler to reason about.

### DIFF
--- a/test/local/email_record_tests.js
+++ b/test/local/email_record_tests.js
@@ -216,9 +216,9 @@ test(
     er.addHit()
     er.addHit()
     t.equal(er.shouldBlock(), false, 'account is not blocked')
-    t.equal(er.update('accountCreate'), 0, 'email action above the email limit')
+    t.equal(er.update('accountCreate'), 1, 'email action above the email limit')
     t.equal(er.shouldBlock(), true, 'account is now blocked')
-    t.equal(er.update('accountCreate'), 0, 'email action in a blocked account')
+    t.equal(er.update('accountCreate'), 1, 'email action in a blocked account')
 
     t.equal(er.ub.length, 0, 'no unblock attempts')
     er.update('bogus', true)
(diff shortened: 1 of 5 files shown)


test/local/email_record_tests.js
```diff
@@ -216,9 +216,9 @@ test(
     er.addHit()
     er.addHit()
     t.equal(er.shouldBlock(), false, 'account is not blocked')
-    t.equal(er.update('accountCreate'), 0, 'email action above the email limit')
+    t.equal(er.update('accountCreate'), 1, 'email action above the email limit')
     t.equal(er.shouldBlock(), true, 'account is now blocked')
-    t.equal(er.update('accountCreate'), 0, 'email action in a blocked account')
+    t.equal(er.update('accountCreate'), 1, 'email action in a blocked account')
```
rfk: We can see here that several tests have hacked around this problem in the past.  The comment says we're asserting that an action was blocked, but `er.update` is returning zero which is supposed to indicate success.

Issue #160, opened 2017-01-30
Support rating-limit on SMS actions


vbudhram:
In support of https://github.com/mozilla/fxa-features/issues/53, this issue tracks adding rate limiting to SMS actions. Initial it will support the following use cases:

* Rate limit sending to a specific SMS, X amount of times
* Rate limit sending X SMS from an IP address

PR #161, opened 2017-01-31
feat(sms): Add support for rate-limiting sms actions

vbudhram:
Fixes https://github.com/mozilla/fxa-customs-server/issues/160

This PR adds the following support to the custom's server:

- [x] Rate limit sending X amount of sms to a specific number
- [x] Rate limit an ip sending X total amount of sms

Connects with https://github.com/mozilla/fxa-features/issues/53, https://github.com/mozilla/fxa-auth-server/issues/1628

### DIFF
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -43,6 +43,12 @@ var EMAIL_SENDING_ACTION = {
   sendUnblockCode: true
 }
 
+// Actions that can send sms, and could make us
+// very annoying to a user if abused.
+var SMS_SENDING_ACTION = {
+  inviteUserSMS: true
+}
+
 module.exports = {
 
   isPasswordCheckingAction: function isPasswordCheckingAction(action) {
@@ -59,6 +65,9 @@ module.exports = {
 
   isEmailSendingAction: function isEmailSendingAction(action) {
     return EMAIL_SENDING_ACTION[action]
-  }
+  },
 
+  isSMSSendingAction: function isSMSSendingAction(action) {
+    return SMS_SENDING_ACTION[action]
+  }
 }
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -46,7 +46,7 @@ const EMAIL_SENDING_ACTION = {
 // Actions that can send sms, and could make us
 // very annoying to a user if abused.
 const SMS_SENDING_ACTION = {
-  inviteUserSms: true
+  connectDeviceSms: true
 }
 
 module.exports = {
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -4,7 +4,7 @@
 
 // Actions that, if allowed, would allow an attacker
 // to try a candidtate password against an account.
-var PASSWORD_CHECKING_ACTION = {
+const PASSWORD_CHECKING_ACTION = {
   accountLogin: true,
   accountDestroy: true,
   passwordChange: true
@@ -14,7 +14,7 @@ var PASSWORD_CHECKING_ACTION = {
 // to try a guess at a randomly-generated security code.
 // Code are higher entropy so we can allow more of these,
 // but if you're doing it a lot, you're probably a baddie.
-var CODE_VERIFYING_ACTION = {
+const CODE_VERIFYING_ACTION = {
   recoveryEmailVerifyCode: true,
   passwordForgotVerifyCode: true
 }
@@ -23,7 +23,7 @@ var CODE_VERIFYING_ACTION = {
 // to check whether an account exists for a particular user.
 // Basically any unauthenticated endpoint that takes
 // an email address as input.
-var ACCOUNT_STATUS_ACTION = {
+const ACCOUNT_STATUS_ACTION = {
   accountCreate: true,
   accountLogin: true,
   accountDestroy: true,
@@ -35,7 +35,7 @@ var ACCOUNT_STATUS_ACTION = {
 
 // Actions that send an email, and hence might make
 // us look like spammers if abused.
-var EMAIL_SENDING_ACTION = {
+const EMAIL_SENDING_ACTION = {
   accountCreate: true,
   recoveryEmailResendCode: true,
   passwordForgotSendCode: true,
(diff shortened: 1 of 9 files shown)


vbudhram:
@philbooth @rfk This provides the rate limiting functionality for sms requests. r?

rfk:
@vbudhram this looks solid at a first glance, hopefully I can find the time for a thorough review soon.

But you're right that it also really highlights the need for some serious refactoring in this repo!  Let's chat next week about putting a plan in place to untangle the web of logic we've woven here over the years.

lib/config/config.js
```diff
@@ -107,6 +107,20 @@ module.exports = function (fs, path, url, convict) {
...
         env: 'IP_RATE_LIMIT_BAN_DURATION_SECONDS'
       },
+      smsRateLimit: {
+        limitIntervalSeconds: {
+          doc: 'Duration of automatic throttling for sms',
+          default: 60 * 15,
+          format: 'nat',
+          env: 'SMS_RATE_LIMIT_INTERVAL_SECONDS'
+        },
+        maxSMSs: {
```
philbooth: I'm not a fan of the capitalisation here, especially as it's all lower case above. I suspect you did it because the pluralised double-s looks funny in lower case, but maybe that's a reason to ditch the plural. `maxSms`?

lib/actions.js
```diff
@@ -59,6 +65,9 @@ module.exports = {
 
   isEmailSendingAction: function isEmailSendingAction(action) {
     return EMAIL_SENDING_ACTION[action]
-  }
+  },
 
+  isSMSSendingAction: function isSMSSendingAction(action) {
```
philbooth: If you switched to ES6 style here, we could end the senseless tyranny of double-named functions that's been blighting this module.

lib/ip_record.js
```diff
@@ -123,6 +124,32 @@ module.exports = function (limits, now) {
     this.as = this._trim(now, this.as, limits.maxAccountStatusCheck)
   }
 
+  IpRecord.prototype.isOverSMSLimit = function () {
+    this.trimSMSRequests(now())
```
philbooth: ...oh you've capitalised them here as well. I don't know, feel free to ignore what I said above in that case, just as long as it's upper-cased everywhere. A sometimes-uppercase-but-sometimes-lowercase mish-mash would keep me awake at night.

lib/ip_record.js
```diff
@@ -123,6 +124,32 @@ module.exports = function (limits, now) {
...
+    // Limit based on number of unique sms request sent by this IP
+    var count = 0
+    var seen = {}
+    this.sms.forEach(function(info) {
```
philbooth: ...but then I see this little guy and I think, no, all-caps is madness.

lib/ip_record.js
```diff
@@ -123,6 +124,32 @@ module.exports = function (limits, now) {
...
+      if (!(info.u in seen)) {
```
philbooth: I can see they're used elsewhere here but, for the record, I'm not a fan of these one-character (and two-character) property names. It's like we're explicitly trying to make the code harder to follow. Using `t` instead of `time` at least has an element of logic, but why `u` for the phone number? Am I missing something?

lib/server.js
```diff
@@ -77,14 +78,21 @@ module.exports = function createServer(config, log) {
...
       mc.getAsync(email).then(EmailRecord.parse, EmailRecord.parse),
       mc.getAsync(ip).then(IpRecord.parse, IpRecord.parse),
       mc.getAsync(ip + email).then(IpEmailRecord.parse, IpEmailRecord.parse),
       reputationService.get(ip)
     ]
+
+    // Check against SMS records to make sure that this request
+    // can send SMS
+    if (smsNumber) {
+      promises.push(mc.getAsync(smsNumber).then(smsRecord.parse, smsRecord.parse))
```
philbooth: *Scrolls up to the top of the file to look for what `mc` means.*

lib/sms_record.js
```diff
@@ -0,0 +1,85 @@
...
+
+  function SMSRecord() {
+    this.xs = []
+  }
+
+  SMSRecord.parse = function (object) {
+    var rec = new SMSRecord()
+    object = object || {}
+    rec.rl = object.rl       // timestamp when the account was rate-limited
+    rec.xs = object.xs || [] // timestamps when sms were sent
```
philbooth: *Sigh*

philbooth:
Looks fine to me @vbudhram, although I'm not very familiar with this repo so that probably doesn't qualify as an official r+.

This is not aimed at you, but the variable/property names in this repo are *completely mad*, aren't they?

lib/config/config.js
vbudhram: Heh guess a pattern that I thought made sense to me was to capitalize acronyms only if in between words. No issues changing to `maxSmss`.

lib/ip_record.js
vbudhram: Yea I agree, it is unnecessarily hard to read. Had to use `info.u` because I was reusing the logic for [trimming](https://github.com/mozilla/fxa-customs-server/blob/606255fd5e74f2f13cf236fbf2613fcb7cb9a5dc/lib/ip_record.js#L162) hits in the record. This being said, I created the following issue https://github.com/mozilla/fxa-customs-server/issues/163 so that customs server gets some more ♥️.

lib/sms_record.js
vbudhram: @rfk Looking back at code, these records get read/written to memcache. Is that why the code is using as small a possible var names? This pattern was before my time.

lib/actions.js
vbudhram: Removed the extra function names but couldn't go full ES6, https://github.com/mozilla/fxa-customs-server/issues/164.

lib/sms_record.js
rfk: Yes, IIRC exactly this, those properties go straight into memcache as JSON so this was designed to keep the memory pressure down.

rfk: (which is not to argue that it's the right tradeoff *today*, just to explain why it's there)

lib/actions.js
```diff
@@ -35,30 +35,39 @@ var ACCOUNT_STATUS_ACTION = {
...
   recoveryEmailResendCode: true,
   passwordForgotSendCode: true,
   passwordForgotResendCode: true,
   sendUnblockCode: true
 }
 
+// Actions that can send sms, and could make us
+// very annoying to a user if abused.
+const SMS_SENDING_ACTION = {
+  inviteUserSms: true
```
vladikoff: what's the logic behind this `inviteUserSms` name?

vbudhram: I should have made a note about this. Since the auth-server sms endpoint will be generic, I needed some way to differentiate the type of sms being sent. Currently, this should only be for inviting users, however I could see us reusing the sms endpoint for 2FA etc. I am open for other another name.

vbudhram: @vladikoff Thought about it some more and really what is happening is the user is sending an sms to connect a device, not inviting a user. Renamed to `connectDeviceSms`.

vladikoff:
Fixes https://github.com/mozilla/fxa-customs-server/issues/160

lib/ip_record.js
rfk: FWIW the `u` in other places here was for "user".

lib/sms_record.js
rfk: OK, I'll admit, I have no idea what "xs" is supposes to stand for in the email_record class, nor here.  But at least it's consistent between the two I suppose...

rfk:
r+, but with the caveat that we're going to get cracking on that long-overdue refactor in this repo, as discussed IRL earlier today :-)

vbudhram:
@rfk @philbooth Thanks guys!

shane-tomlinson:
🥇

lib/actions.js
```diff
@@ -35,30 +35,39 @@ var ACCOUNT_STATUS_ACTION = {
...
 
+// Actions that can send sms, and could make us
+// very annoying to a user if abused.
+const SMS_SENDING_ACTION = {
+  connectDeviceSms: true
+}
+
 module.exports = {
 
-  isPasswordCheckingAction: function isPasswordCheckingAction(action) {
```
vladikoff: this was here because it is useful during debugging / testing, but we can live without it I hope

vbudhram: I think it will be ok? I don't mind adding it back if we need it.

PR #167, opened 2017-02-06
feat(blocklist): Add support to specify the block level for multiple blocklists

### LINKED ISSUE #165: Make ip blocklist configurable to block on select lists
Currently, if an ip address is found in any blocklist, that request is blocked. We should have the ability to specify what action happens on a blocklist level.

vbudhram:
I took a jab at how I thought this could be implemented. Since convict does not allow an array of objects as properties, we keep a separate list that should be log only.

Fixes #167 

@rfk Thoughts?


lib/ip_blocklist_manager.js
```diff
@@ -31,12 +31,49 @@ module.exports = function (log, config) {
...
+      const listOverrideRules = config.ipBlocklist.listsOverrideRules
+      if (listOverrideRules && listOverrideRules.length > 0 &&
+        self.loadedLists.length === listOverrideRules.length) {
+
+        // Lookup the overrided blocklist config and block
+        // request if set to `true`
+        return blockInfo.listPaths.some(function (listPath) {
+          const listIndex = self.loadedLists.indexOf(listPath)
+
+          return (listOverrideRules[listIndex] === 'true')
```
vbudhram: Not ideal, but from https://github.com/mozilla/node-convict/blob/master/lib/convict.js#L327, the passed array properties will be stored as `string` and not `boolean`.

lib/ip_blocklist.js
```diff
@@ -146,6 +146,10 @@ module.exports = function (log, config) {
 
         return result
       })
+
+      return {
+        found: found
+      }
```
rfk: I don't really understand the change to return an object here - is it just for symmetry with the same method in the manager class?

lib/config/config.js
```diff
@@ -200,11 +200,17 @@ module.exports = function (fs, path, url, convict) {
...
         format: Array,
         default: [],
         env: 'IP_BLOCKLIST_FILES'
       },
+      listsOverrideRules: {
+        doc: 'An array of true/false values that map directly the the lists property and determines if that specific list should be blocked',
+        format: Array,
+        default: [],
+        env: 'IP_BLOCKLIST_FILES_BLOCKING'
+      },
```
rfk: This approach will work, but it seems to make the code hard to follow, and it's hard to think of a good name for the config option that explains how to use it.  I wonder if it would be simpler to just have to separate lists of files, one that is log-only and one that actually blocks, something like:

```
ipBlocklist: {
  enable: true,
  lists: ["/this/one/blocks.netset"],
  logOnlyLists: ["/this/one/logs.netset"]
}
```

vbudhram: Your approach seems much more reasonable and cleaner. I will update accordingly.

lib/ip_blocklist.js
vbudhram: Yea, it was to keep symmetry since they have similar interface and made testing much easier. With change from above, I might not need this anymore.

vbudhram:
@rfk Your recommendation did a pretty good job of simplifying logic. As a side, created [issue] to track the changes needed by puppet from this PR. Mind another r?

lib/server.js
```diff
@@ -22,7 +22,7 @@ module.exports = function createServer(config, log) {
   if (config.ipBlocklist.enable) {
     var IPBlocklistManager = require('./ip_blocklist_manager')(log, config)
     var blockListManager = new IPBlocklistManager()
-    startupDefers.push(blockListManager.load(config.ipBlocklist.lists))
+    startupDefers.push(blockListManager.load([].concat(config.ipBlocklist.lists, config.ipBlocklist.logOnlyLists)))
```
rfk: It seems convoluted to concat the two lists here, and then check `config.ipBlocklist.logOnlyLists` again to distinguish them when we're looking for hits.  I wonder if it'd be cleaner to pass the lists in separately and let the `IPBlocklistManager` class deal with them internally.  Something like:

```
  IPBlocklistManager.prototype.load = function (lists, logOnlyLists) {
    var self = this

    if (! Array.isArray(lists) || ! Array.isArray(logOnlyLists)) {
      throw Error('lists must be an array')
    }

    var loadedLists = lists.map(function (listPath) {
      var blocklist = new IPBlocklist()
      self.ipBlocklists.push(blocklist)
      return blocklist.load(listPath)
    })

    loadedLists = loadedLists.concat(logOnlyLists.map(function (listPath) {
      var blocklist = new IPBlocklist()
      blocklist.logOnly = true
      self.ipBlocklists.push(blocklist)
      return blocklist.load(listPath)
    })

    return Promise.all(loadedLists)
  }
```

In the `contains` method, you could then check the `logOnly` property of each list as you iterate through them, rather than doing a separate `logOnlyLists.indexOf` lookup.

vbudhram:
@rfk Made changes recommended, r?

test/remote/ip_blocklist_tests.js
```diff
@@ -92,6 +96,30 @@ test(
...
+    )
+  }
+)
+
+test(
+  'block from hit on logOnly list and blocklist',
+  function (t) {
+    client.postAsync('/check', {ip: LOG_ONLY_BOTH_LIST_IP, email: TEST_EMAIL, action: ACTION},
+      function (err, req, res, obj) {
+        t.equal(obj.block, true, 'request is not blocked')
```
rfk: nit: should be "request is blocked"

PR #169, opened 2017-02-08
fix(startup): Exit process on any startup error.

vbudhram:
Fixes an issue where the custom server process didn't exit properly on startup errors.

### DIFF
--- a/bin/customs_server.js
+++ b/bin/customs_server.js
@@ -12,8 +12,8 @@ var server = require('../lib/server')
 var config = require('../lib/config').getProperties()
 var log = require('../lib/log')(config.log.level, 'customs-server')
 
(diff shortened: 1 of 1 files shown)


vbudhram:
@jrgm r?

bin/customs_server.js
```diff
@@ -32,4 +32,5 @@ server(config, log)
   })
   .catch(function (err) {
     log.error({op: 'customs.bin.error', err: err})
+    shutdown()
```
rfk: nit: please add an exit-code argument to `shutdown` and pass it to `process.exit`, then call `shutdown(1)` here.  Otherwise the process will appear to have exited successfully.

PR #170, opened 2017-02-08
fix(timers): Unref all the update-polling timers.

rfk:
These timers are all just for background polling for updates, they don't need to keep the event-loop alive.  Unref them.


rfk:
I think a small constant number should be OK, but if you create an unref'd timer for every request then you might get into trouble.

PR #171, opened 2017-02-09
fix(config-set): Don't attempt to merge array-valued config items.

### LINKED ISSUE #168: Oddity with PR 151; handling of JSON array merging
Start with local memcache with no data (command `flush_all`). Then do this
```
File ary.json:

{
  "allowedIPs": [ 
    "55.55.55.55"
  ]
}

$ ./scripts/config-list.js 127.0.0.1:11211; \
    cat /tmp/ary.json | ./scripts/config-set.js 127.0.0.1:11211; \
    echo @@@@@; ./scripts/config-list.js 127.0.0.1:11211
no limits set
no allowedIPs set
no allowedEmailDomains set
no requestChecks set
{
  "limits": {},
  "allowedIPs": [],
  "allowedEmailDomains": [],
  "requestChecks": {}
}
@@@@@
no limits set
no allowedEmailDomains set
no requestChecks set
{
  "limits": {},
  "allowedIPs": {
    "0": "55.55.55.55"
  },
  "allowedEmailDomains": [],
  "requestChecks": {}
}
```

So `allowedIPs` comes back as a hash with numeric keys. This was not the previous behaviour. (And yeah, an Array is an Object with numeric keys and a .length, but....).

rfk:
Fixes #168; @jrgm r?

### DIFF
--- a/scripts/config-set.js
+++ b/scripts/config-set.js
@@ -23,6 +23,7 @@ var Memcached = require('memcached')
 var P = require('bluebird')
 var BL = require('bl')
 var merge = require('lodash.merge')
+var isEqual = require('lodash.isequal')
 
 P.promisifyAll(Memcached.prototype)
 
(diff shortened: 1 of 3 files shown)


scripts/config-set.js
```diff
@@ -47,6 +48,20 @@ function writeMergedSettings(mc, key, newSettings) {
     )
 }
 
+function clobberSettings(mc, key, newSettings) {
```
jrgm: Could throw in:
```
+  if (! Array.isArray(newSettings)) {
+    console.error('newSettings is not an Array', newSettings)
+  }
```
FWIW. Type checking "hashes" is left as an exercise to the reader.

Issue #166, opened 2017-02-04
config option reputationService.host must be able to accept a FQDN value


jrgm:
1) In stage/production, the tigerblood endpoint is set up to use an ELB. ELB's advertise multiple IP's and those IP's can change at anytime. 

2) Current master of `ip-reputation-js-client`, not yet published to npm, hardcodes the scheme to `https`, which cannot be used with IP addresses.

jrgm:
See also https://github.com/mozilla-services/ip-reputation-js-client/issues/3, where the client should be strictSSL and not follow any redirects. (Not directly related to this bug, but both this repo and that repo need to align).

vbudhram:
Spoke with @g-k and he said that he was working on some patches for this, possibly sending the service url instead.

Connects with https://github.com/mozilla-services/ip-reputation-js-client/issues/3 and https://github.com/mozilla-services/ip-reputation-js-client/pull/4

g-k:
ugh, yeah.

The package is published: https://www.npmjs.com/package/ip-reputation-js-client.

If https://github.com/mozilla-services/ip-reputation-js-client/pull/4 addresses https://github.com/mozilla-services/ip-reputation-js-client/issues/3, then I need to:

- [ ] publish a new major version (constructor options changed)
- [ ] update package.json (and shrinkwrap?) on https://github.com/mozilla/fxa-customs-server/compare/master...g-k:fix-rep-service-config?expand=1
- [ ] land that branch if it looks OK
- [ ] update the stage and prod fxa puppet config
  - [ ] remove `fxa::auth_customs_server::reputation_service_ip` 
  - [ ] add `fxa::auth_customs_server::reputation_service_base_url` set to`https://tigerblood.{stage|prod}.mozaws.net`
  - [ ] re-enable in stage (revert https://github.com/mozilla-services/puppet-config/pull/2476)

vladikoff:
This was corrected...

PR #172, opened 2017-02-10
Fix rep service config

g-k:
Refs: https://github.com/mozilla/fxa-customs-server/issues/166

Updates ip rep js client to 2.0.1 and makes the relevant test and server changes.

I didn't update the shrinkwrap.  Is that part of the release process or should it be in the PR?


jrgm:
LGTM.

Issue #128, opened 2016-08-24
A modest proposal for a major refactor


rfk:
I've been spending a lot of background cycles thinking about the structure of code in this repo, and talking though some parts of it with @seanmonstar.  I think the way it's currently structured is hurting us in a variety of ways:
- Many of the rules implemented here are ad-hoc, and it's not always obvious how particular config values map to particular concrete effects.  See, for example, the recent change w.r.t. blocking email-sending actions at the IP level.
- We want to hide certain implementation details of our rules, e.g. various well-formedness checks that we apply to login requests.  That currently means deploying from a private fork of this repo, because those checks have to be rendered in code.
- It's hard to emit detailed metrics about what's being blocked and why; I've seen two aborted attempts at https://github.com/mozilla/fxa-customs-server/issues/106 because it would involve major surgery to the structure of the code.

Given that we now have significant operational experience with this repo, it's time to consider refactoring it to make the rules it offers easier to understand, easier to change and configure, and easier to monitor in production.

So, a proposal based around a small set of unifying concepts:
- The whole job of this server is to check an incoming **request** and make recommendations on what to do with it.  The data we have about the request includes the source ip address, the target email or uid, the target action, the headers, body payload, etc.  We basically have this structure right now, although it could be made more explicit in the code.
- We check said **request** by applying a list of **rules**.  Each rule has a name and can be independently configured.  Each rule is an instance of one of a small number of concrete _types_ of rule, such as "rate limit" or "check request structure", for which we provide implementations in the public repo.  But the precise details of each instantiation of a rule come from private config.
- Whenever a rule is matched, we can apply one or more **sanctions** to the request.  We might recommend it be blocked, or rate-limited, or treated with suspicion.  We might add the IP or the email to an internal blocklist.  We have a small set of these available and they're orthogonal to the list of rules.

Having an explicit list of rules with meaningful names, will let us log much better metrics about which of them are triggering and against what requests.  It will also make it easier to understand exactly what the rules _are_ at any point in time.  And having them be instantiated from config will mean we can keep various details private just by putting them in the puppet-config repo and/or directly in memcached.

What might it look like in practice?  Here's a strawman config that could replace some of the current hard-coded rules with something more declarative.  (Don't get too caught up in the details, I'm just trying to get the idea out of my head).

```
{
  // Some global config that's the same as we have right now
  "allowedIps": ["63.245.214.162", "63.245.214.168" ],
  "allowedEmaiLDomains": ["restmail.net"],

[...]

PR #174, opened 2017-02-14
WIP: Refactor all the things.  All of them.

rfk:
Connects to #128.  I'm opening this for visibility, but it's nowhere near ready.  Like, it's half-backed and doesn't even run.  It's going to be my scratch-pad for some Friday experiments in refactoring this service, disentangling our many rules into something that's easier to understand.


lib/config.js
```diff
@@ -0,0 +1,286 @@
...
+        },
+
+        limit_logins_daily: {
+          type: 'RateLimit',
+          action: 'checkPassword',
+          config: {
+            limit: 30,
+            period: '1 day'
+          }
+        },
```
rfk: Here, I'm trying to re-imagine our rules as an explicit set of named rules that we can look and and understand.  They'll probably end up being slightly different to what's currently codified in code, but that seems OK as long as we're still happy they provide equivalent protection.

lib/judgement.js
```diff
@@ -0,0 +1,40 @@
...
+
+function Judgement(options) {
+  this.triggeredRules = options.triggeredRules || []
+  this.suspect = options.suspect || false
+  this.block = options.block || false
+  this.retryAfter = options.retryAfter || 0
+  this.canUnblock = options.canUnblock || true
+}
+
+Judgement.prototype.include = function include(rule, sanctions, record) {
```
rfk: I'm aware this is a very hokey name, I'm just trying to tease apart the different pieces for now...

lib/judgement.js
```diff
@@ -0,0 +1,40 @@
...
+  this.triggeredRules.push(rule.name)
```
rfk: Note here that when we return our judgement on a request, we include the list of rules that were triggered to produce it.  Very important for metrics purposes.

lib/rules/ratelimitunique.js
```diff
@@ -0,0 +1,74 @@
...
+        n++
+      }
+      hit = state.hits[--i]
+      if (i === -1) {
+        break
+      }
+    }
+    state.hits = state.hits.slice(i + 1)
+  }
+}
```
rfk: Imagine the simple joy of being able to re-use this logic from a single place, rather than copy-pasting it all over the place :-)

rfk:
This doesn't need to hang around being open

PR #176, opened 2017-02-15
feat(docker): add Docker support

seanmonstar:


docker-compose.yml
```diff
@@ -0,0 +1,10 @@
+version: '2'
```
jbuck: I'd agree with your/Ryans assertion that we can delete for this now.

Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM ubuntu:trusty
```
vladikoff: 👍

seanmonstar:
@jbuck ok, updated. Added a circle.yml, and added integration. I'm noticing that the build is failing cause it's not published to the docker library. Is that required to run the tests? I'd think that running this on circle would use the code in the repo, not something already published...

Dockerfile-test
```diff
@@ -0,0 +1,2 @@
+FROM fxa-custom-server:build
```
jbuck: The reason it's trying to pull from docker hub is the missing `s` here, should be `fxa-customs-server`

seanmonstar:
its working!!

Dockerfile-build
```diff
@@ -0,0 +1,14 @@
+FROM mhart/alpine-node:4.7.2
```
jbuck: Lets update this to 4.8.0 since that's out now

seanmonstar:
Updated the alpine-node version to 4.8.0.

PR #179, opened 2017-02-21
fix(reputation): add more (positive) logging to reputation service requests

jrgm:
r? - @vbudhram, @g-k 

Add some more logging, so 200 vs. 400 responses can be seen in logs. 

May make more sense to review with `?w=1` due to whitespace changes.


jrgm:
https://github.com/mozilla/fxa-customs-server/pull/179/files?w=1

PR #181, opened 2017-02-24
chore(package): Use ip-rep service client with keepalive enabled

g-k:
Bump IP Reputation JS client to version 2.1.1 which has keepalive enabled.

This should help fix some of the TCP RSTs and timeouts jrgm and I saw when we enabled a canary instance in prod last week.


vladikoff:
Looks good, @g-k this is set to be tagged for Train-82 at the moment

PR #191, opened 2017-03-27
fix(tests): Update rate limit sms by ip address

### LINKED ISSUE #190: SMS throttled limit is not honored when SMSs are send to combination of same & different phone numbers
Prerequisite: Existing SessionToken
Two different mobile numbers
STR:

1. Navigate to 'https://accounts.firefox.com/sms?service=sync'
2. Send 3 SMSs to phone number #1
3. Send 2 SMSs to phone number #2
4. Send 6 th SMS to phone number #1

Expected: 6th SMS should get throttled. SMS throttled limit of 5 messages should be honored for a specific FxA account even on same machine

Actual: 6th SMS successfully sent to mobile device. 

Note: This happens with any combination of phone numbers. e.g. send sms to 3 different numbers and then 3 to same phone numbers. 
Throttled limit only work either with 5 SMSs to same phone number or to 5 different phone numbers or if max SMSs (5) limit is reached on one phone number (in this case, 5 SMSs to phone number #1). 
So in above case If I send 2 more SMSs to mobile # 1 after step 4 then  for 8 th SMS would get throttled.

vbudhram:
This PR correctly rate limits when a user has sent more than 5 sms from an ip address. Previously, the the rate limiting was `rate-limit when 5 unique sms requests sent from this ip address`. Updated tests to be more reflective of this scenario.

Fixes #190

@rfk r?

### DIFF
--- a/lib/ip_record.js
+++ b/lib/ip_record.js
@@ -126,28 +126,28 @@ module.exports = function (limits, now) {
 
   IpRecord.prototype.isOverSmsLimit = function () {
     this.trimSmsRequests(now())
-    // Limit based on number of unique sms request sent by this IP
-    var count = 0
-    var seen = {}
-    this.sms.forEach(function(info) {
-      if (!(info.u in seen)) {
-        count += 1
-        seen[info.u] = true
-      }
-    })
-    return count > limits.maxSms
+    return this.sms.length > limits.maxSms
   }
 
-  IpRecord.prototype.addSmsRequest = function (info) {
-    info = info || {}
-    var t = now()
-    var phoneNumber = info.phoneNumber || ''
-    this.trimSmsRequests(t)
-    this.sms.push({ t: t, u: phoneNumber })
+  IpRecord.prototype.addSmsRequest = function () {
+    this.sms.push(now())
   }
 
   IpRecord.prototype.trimSmsRequests = function (now) {
-    this.sms = this._trim(now, this.sms, limits.maxSms)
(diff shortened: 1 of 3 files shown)


lib/ip_record.js
```diff
@@ -126,28 +126,26 @@ module.exports = function (limits, now) {
...
-    this.sms.push({ t: t, u: phoneNumber })
+  IpRecord.prototype.addSmsRequest = function () {
+    this.sms.push(now())
   }
 
   IpRecord.prototype.trimSmsRequests = function (now) {
-    this.sms = this._trim(now, this.sms, limits.maxSms)
+    if (this.sms.length === 0) { return }
+    // xs is naturally ordered from oldest to newest
+    // and we only need to keep up to limits.maxEmails + 1
```
rfk: comment nit: s/maxEmails/maxSms/

lib/ip_record.js
```diff
@@ -236,8 +234,8 @@ module.exports = function (limits, now) {
     }
 
     // Increment sms request count and throttle if needed
-    if (actions.isSmsSendingAction(action) && phoneNumber) {
-      this.addSmsRequest({ phoneNumber: phoneNumber })
```
rfk: The `phoneNumber` argument appears to be unused after removing these, can we safely remote it or does it need to be kept for consistency with other records?

vbudhram: Good catch, it can be removed.

PR #197, opened 2017-04-19
WIP refactor(check): Extract `updateBlocks` from the `/check` route.

### LINKED ISSUE #195: log why users are blocked in call to /check
/check logs the action the check is being done for in `action`, but does not log which rule the check failed. This info would be useful to determine the reason for the recent uptick in blocked signins.

shane-tomlinson:
This is the beginning of the work to log a "reason" whenever a block
occurs.

Fixes #195

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -128,6 +128,27 @@ module.exports = function createServer(config, log) {
     return rawEmail.toLowerCase()
   }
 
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
Killing this PR, I'm not going to get to this!

PR #198, opened 2017-04-20
fix(sms): Add ability to rate-limit sms by email

### LINKED ISSUE #188: SMS throttled limit is not honored for a specific FxA account if different machines are used for sending SMSs
Prerequisite:
1. Existing SessionToken
2. Availability of multiple machines

STR:
1. On a specific machine, navigate to 'https://accounts.firefox.com/sms?service=sync'
2. Enter phone number in mobile number field and click send to send SMS.
3. Repeat step 1 and 2 for 4 times to send 4 messages to valid phone numbers within quick span (less than an hour).
4. Using the same FxA account repeat step 1, 2, & 3 on multiple machines (same office wifi/IP Address) .

Expected: SMS throttled limit of 5 messages should be honored for a specific FxA account even if different machines (same office wifi/IP Address) are used for sending SMSs.

Actual: SMS throttled limit is not honored for a specific FxA account if different machines (same office wifi/IP Address) are used for sending SMSs.

For e.g. It is very much possible to send up to 40 SMSs from the same FxA account if the user tries from 10 different systems.

vbudhram:
This PR adds ability to rate limit SMS actions by email.

Fixes #188 

@rfk @seanmonstar r?

### DIFF
--- a/lib/email_record.js
+++ b/lib/email_record.js
@@ -91,7 +91,7 @@ module.exports = function (limits, now) {
 
   EmailRecord.prototype.trimSmsRequests = function (now) {
     if (this.sms.length === 0) { return }
-    // vc is naturally ordered from oldest to newest
+    // sms is naturally ordered from oldest to newest
     // and we only need to keep up to limits.maxSms + 1
 
     var i = this.sms.length - 1
--- a/lib/email_record.js
+++ b/lib/email_record.js
@@ -12,6 +12,7 @@ module.exports = function (limits, now) {
   function EmailRecord() {
     this.vc = []
     this.xs = []
+    this.sms = []
     this.ub = []
   }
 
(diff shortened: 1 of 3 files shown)


lib/email_record.js
```diff
@@ -82,6 +84,30 @@ module.exports = function (limits, now) {
...
   }
 
+  EmailRecord.prototype.isOverSmsLimit = function () {
+    this.trimSmsRequests(now())
+    return this.sms.length > limits.maxSms
+  }
+
+  EmailRecord.prototype.trimSmsRequests = function (now) {
+    if (this.sms.length === 0) { return }
+    // vc is naturally ordered from oldest to newest
```
rfk: s/vc/sms/ in this comment

vbudhram:
Thanks @rfk!

PR #200, opened 2017-04-20
fix(settings): pushOnMissing no longer updates on unexpected errors

### LINKED ISSUE #194: adjust the Limits pushOnEmpty to not consume unrelated errors

seanmonstar:
This started as just adjusting the `pushOnMissing` code in `limits.js`, and then I noticed that this logic existed in several places. Instead of just injecting a fix into each of those locations, I've refactored all of this logic into a common `Settings` class.

The actual fix to the issue is that we will now only try to push if there was an explicit `PushOnMissing` error thrown, otherwise it's propagated back out, including a `log.error` message.

Closes #194


rfk:
> Instead of just injecting a fix into each of those locations, I've refactored all of this logic into a common Settings class.

You are a gentleman and a scholar, thank you!

lib/settings/settings.js
```diff
@@ -0,0 +1,85 @@
...
+
+const KEY = Symbol()
+const POLL_INTERVAL = Symbol()
+
+module.exports = (config, mc, log) => {
+
+
+  // A sentinel error for signaling that the result was invalid/missing,
+  // and we should try to push our own representation to memcached.
+  class PushOnMissing extends Error {}
```
rfk: api nit: something about the name `PushOnMissing` here doesn't feel right to me.  The correspondingly-named argument to `refresh` means "you should push if the item turns out to be missing", the situation here is "the item was actually missing".  Would naming this something like `ItemMissing` make things clearer?

I don't feel super strongly about it though.

lib/settings/settings.js
```diff
@@ -0,0 +1,85 @@
...
+        .then(() => this.refresh())
+    }
+
+    refresh(options) {
+      log.info({ op: this[KEY] + '.refresh' })
+      let result = mc.getAsync(this[KEY]).then(value => this.validate(value))
+
+      if (options && options.pushOnMissing) {
+        result = result.catch(err => {
+          if (err instanceof PushOnMissing) {
```
rfk: IIUC, the "if missing" part of this kind of happens implicitly:

* We try to the value from memcache, but it's missing, so we get `null` or `undefined` or something
* We call `this.validate(value)` with the missing result
* The implementation of `this.validate` does some type-checking and errors out with `PushOnMissing` because it's not an arrow/object/what-have-you

Is it worth adding an explicit check for "the item was missing" here in the base class, in case a future subclass doesn't know that it have to perform this check inside `validate()`?

lib/settings/settings.js
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    pollForUpdates() {
+      this.stopPolling()
+      this[POLL_INTERVAL] = setInterval(this.refresh.bind(this), config.updatePollIntervalSeconds * 1000)
+      this[POLL_INTERVAL].unref()
+    }
+
+    stopPolling() {
+      clearInterval(this[POLL_INTERVAL])
```
rfk: The use of symbol-valued properties here took me a while to get my head around, but I can't suggest a better approach.  It may be worth a comment at the point where the symbols are defined, pointing out why we're using them (unless this is a totally normal thing and I just haven't paid enough attention to latest javascript idioms to notice it...)

lib/settings/settings.js
```diff
@@ -0,0 +1,85 @@
...
+        err => {
+          log.error({ op: this[KEY] + '.refresh', err: err })
+          throw err
+        }
+      )
+    }
+
+    pollForUpdates() {
+      this.stopPolling()
+      this[POLL_INTERVAL] = setInterval(this.refresh.bind(this), config.updatePollIntervalSeconds * 1000)
```
rfk: nit: use a fat-arrow here

lib/settings/settings.js
```diff
@@ -0,0 +1,85 @@
...
+          } else {
+            throw err
+          }
+        })
+      }
+      return result.then(
+        value => this.setAll(value),
+        err => {
+          log.error({ op: this[KEY] + '.refresh', err: err })
+          throw err
```
rfk: This potentially-uncaught error won't, like, bring down the whole process or anything will it? :-)

lib/settings/settings.js
seanmonstar: It's the JavaScript way of private properties. The nodejs source has been growing their use of `Symbol` for any new private properties they need (can't really remove all the `_prop`, even though the underscore was supposed stop people from using, but lol).

There was 2 reasons to use these here:

1. These properties should be private, there's no reason for a sub class to fiddling with them.
2. Properties set by a `Symbol` aren't enumerable, so they don't get serialized into JSON. This could be done with `Object.setProperty`, but `Symbol` seemed better.

Farther Future Us should hopefully be quite used to `Symbol` in nodejs code, but I can add a comment for Short Future Us if that seems valuable.

lib/settings/settings.js
seanmonstar: I read this over a few times, and couldn't grasp what you were getting at :_(.

My guess is that you mean the default implementation of `Settings.validate` is just the identity function, so if this wasn't overridden, we would end up passing `this.setAll(null)`, which could cause null errors. If so, I can add a simple null check in `Settings.validate`.

lib/settings/settings.js
seanmonstar: It could, depending on our `unhandledRejection` and `uncaughtException` handlers. Double checking, I see that re-throwing is an addition in this PR, it was previously just logged. I don't have a strong opinion either way (other than swallowing errors makes me sad).

lib/settings/settings.js
rfk: If it's a common thing these days, I'm happy to chalk it up to "TIL" and leave it as-is :-)

lib/settings/settings.js
seanmonstar: To reduce the amount of behavioral changes in this PR, I've reverted to not re-throwing. If we want that, we can explicitly change that in a new PR.

lib/settings/settings.js
seanmonstar: I've added a default implementation of `validate` which throws if the value is null-like.

lib/settings/settings.js
seanmonstar: Actually, the tests were relying on the error to be returned so as to test that other errors besides `Missing` did not trigger a refresh. Instead, I've added `.catch(() => {})` in the poll and in the server start up, saying that we're explicitly ignoring these errors.

lib/settings/settings.js
rfk: > My guess is that you mean the default implementation of Settings.validate
> is just the identity function, so if this wasn't overridden

More specifically, I was worried about the overrider forgetting to check whether the item existed at all.  Having a default implementation seems fine to me, +1.

PR #203, opened 2017-05-18
feat(actions): add consumeSigninCode as an ACCOUNT_ACCESS_ACTION

### LINKED ISSUE #201: Add rate-limiting for signinCodes
Related: mozilla/fxa-auth-server#1882

philbooth:
Fixes #201.

* Adds a new endpoint `/checkIpOnly` that rate-limits purely against an IP address, with no requirement for an email address or a uid.

* Adds a new class of action that I'm calling `ACCOUNT_ACCESS_ACTION`, with one eye on phase 4 of connect-another-device. I needed a new action because I have no email address to rate-limit against, so I can't use `CODE_VERIFYING_ACTION` for instance.

* Refactors some of the common code to make `email` an optional parameter. Doing this necessitated changing the order of some arguments and returned promises, but I think I covered all the calling code in my changes.

The intention is for the proposed auth server endpoint for consuming signinCodes (see mozilla/fxa-auth-server#1882) to hit `/checkIpOnly` with `{"action":"consumeSigninCode"}` and the IP address.

This is the first time I've worked in the customs server and there's a fair bit of stuff I still don't understand, so this PR may have a little way to go. I'll add inline comments to call out the bits I'm most unsure of.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -16,7 +16,8 @@ const PASSWORD_CHECKING_ACTION = {
 // but if you're doing it a lot, you're probably a baddie.
 const CODE_VERIFYING_ACTION = {
   recoveryEmailVerifyCode: true,
-  passwordForgotVerifyCode: true
+  passwordForgotVerifyCode: true,
+  signinCode: true
 }
 
 // Actions that, if allowed, would allow an attacker
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -16,8 +16,7 @@ const PASSWORD_CHECKING_ACTION = {
 // but if you're doing it a lot, you're probably a baddie.
 const CODE_VERIFYING_ACTION = {
   recoveryEmailVerifyCode: true,
-  passwordForgotVerifyCode: true,
-  signinCode: true
+  passwordForgotVerifyCode: true
 }
 
 // Actions that, if allowed, would allow an attacker
@@ -51,6 +50,12 @@ const SMS_SENDING_ACTION = {
   connectDeviceSms: true
 }
 
+// Actions that may grant access to an account but
+// are not associated with an email address or uid.
+const ACCOUNT_ACCESS_ACTION = new Set([
+  'consumeSigninCode'
+])
+
 module.exports = {
 
   isPasswordCheckingAction: function (action) {
@@ -71,5 +76,9 @@ module.exports = {
 
   isSmsSendingAction: function (action) {
     return SMS_SENDING_ACTION[action]
+  },
+
+  isAccountAccessAction (action) {
+    return ACCOUNT_ACCESS_ACTION.has(action)
   }
 }
--- a/lib/ip_record.js
+++ b/lib/ip_record.js
@@ -25,6 +25,7 @@ module.exports = function (limits, now) {
     rec.vc = object.vc || []    // timestamp+email when code verifications occurred
     rec.as = object.as || []    // timestamp+email when account status checks occurred
     rec.sms = object.sms || []  // timestamp+sms when sms sent
+    rec.aa = object.aa || []    // timestamp when account access was attempted
     rec.rl = object.rl          // timestamp when the account was rate-limited
     return rec
   }
@@ -150,6 +151,14 @@ module.exports = function (limits, now) {
     this.sms = this.sms.slice(i + 1)
   }
 
+  IpRecord.prototype.addAccountAccess = function () {
+    this.aa.push(now())
+  }
+
+  IpRecord.prototype.isOverAccountAccessLimit = function () {
+    return this.aa.length > limits.maxAccountAccess
+  }
+
   IpRecord.prototype._trim = function (now, items, maxUnique) {
     if (items.length === 0) { return items }
     // the list is naturally ordered from oldest to newest,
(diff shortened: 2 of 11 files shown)


lib/server.js
```diff
@@ -87,15 +87,19 @@ module.exports = function createServer(config, log) {
     throw err
   }
 
-  function fetchRecords(email, ip, phoneNumber) {
+  function fetchRecords(ip, email, phoneNumber) {
```
philbooth: Argument order changed here to make `email` optional.

lib/server.js
```diff
@@ -87,15 +87,19 @@ module.exports = function createServer(config, log) {
...
       mc.getAsync(ip).then(IpRecord.parse, IpRecord.parse),
-      mc.getAsync(ip + email).then(IpEmailRecord.parse, IpEmailRecord.parse),
       reputationService.get(ip)
     ]
 
+    // The /checkIpOnly endpoint has no email (or phoneNumber)
+    if (email) {
+      promises.push(mc.getAsync(email).then(EmailRecord.parse, EmailRecord.parse))
+      promises.push(mc.getAsync(ip + email).then(IpEmailRecord.parse, IpEmailRecord.parse))
+    }
```
philbooth: ...which obviously means the result order changes too.

lib/server.js
```diff
@@ -110,13 +114,21 @@ module.exports = function createServer(config, log) {
     return mc.setAsync(key, record, lifetime)
   }
 
-  function setRecords(email, ip, emailRecord, ipRecord, ipEmailRecord, phoneNumber, smsRecord) {
-    var promises =       [
-      setRecord(email, emailRecord),
-      setRecord(ip, ipRecord),
-      setRecord(ip + email, ipEmailRecord)
+  function setRecords(ip, ipRecord, email, emailRecord, ipEmailRecord, phoneNumber, smsRecord) {
```
philbooth: Again, argument order changed to move IP-only stuff to the front.

lib/ip_record.js
```diff
@@ -150,6 +151,14 @@ module.exports = function (limits, now) {
     this.sms = this.sms.slice(i + 1)
   }
 
+  IpRecord.prototype.addAccountAccess = function (now) {
+    this.aa.push(now)
+  }
+
+  IpRecord.prototype.isOverAccountAccessLimit = function (now) {
+    return this.aa.length > limits.maxAccountAccess
```
philbooth: Other `isOver...` methods trim their array but I don't see any need to do that here because the array gets truncated in `rateLimit` anyway.

lib/ip_record.js
```diff
@@ -191,6 +200,7 @@ module.exports = function (limits, now) {
     this.rl = now()
     this.as = []
     this.sms = []
+    this.aa = []
```
philbooth: This is the truncation that meant I don't see why I need to trim.

philbooth:
> According to coveralls, I'm missing some test coverage for the new code. Working on it.

Fixed this now. @mozilla/fxa-devs r?

lib/server.js
```diff
@@ -340,6 +352,56 @@ module.exports = function createServer(config, log) {
...
+          block = true
+          retryAfter = ipRecord.retryAfter()
+          blockReason = blockReasons.IP_BAD_REPUTATION
+        }
+
+        return setRecords(ip, ipRecord)
+          .then(() => ({ block, blockReason, retryAfter, suspect }))
+      })
+      .then(result => {
+        res.send(result)
```
vbudhram: I think we should add a check for whitelisted IPs, similar to https://github.com/mozilla/fxa-customs-server/blob/96858c0c028e10bf521fe684120d40d88d1b0553/lib/server.js#L247. Maybe we pull this into some common place for use.

test/remote/ip_blocklist_tests.js
```diff
@@ -59,65 +60,52 @@ var client = restify.createJsonClient({
...
     )
-  }
-)
+  })
 
-test(
-  'should block request on hit form logOnly and blocklist',
-  function (t) {
-    client.postAsync('/check', {ip: LOG_ONLY_BOTH_LIST_IP, email: TEST_EMAIL, action: ACTION},
+  test('should block request on hit form logOnly and blocklist', t => {
```
vbudhram: Typo: should be `from`

lib/actions.js
```diff
@@ -50,6 +50,12 @@ const SMS_SENDING_ACTION = {
   connectDeviceSms: true
 }
 
+// Actions that may grant access to an account but
+// are not associated with an email address or uid.
+const ACCOUNT_ACCESS_ACTION = new Set([
```
vbudhram: 👍 on the name, pretty clear to me what it is doing.

lib/server.js
vbudhram: I think we could add a check for whitelisted ips similar to https://github.com/mozilla/fxa-customs-server/blob/96858c0c028e10bf521fe684120d40d88d1b0553/lib/server.js#L247. Could also pull that into common function for resuse.

lib/actions.js
vbudhram: 👍 on the name, clear and concise.

philbooth:
Thanks for your help @vbudhram, whitelisting stuff and more tests added!

PR #206, opened 2017-06-22
fix(iprecord): clear ip record rate limit after whitelisting

### LINKED ISSUE #205: Current implementation of allowWhitelisted causes problems for /checkIpOnly
When running the content server functional tests, lots of requests to `/check` originate from the same IP address. These cause the rate limit flag to be set in `IpRecord`, but the request itself is not blocked because the email address matches the test in `allowWhitelisted`.

This is all fine until requests to `/checkIpOnly` start turning up. Now, the `IpRecord` is still marked as rate-limited when it is fetched from memcached, but we have no email address to whitelist on. So those requests get blocked, even though the specific action they're checking is below its threshold.

It seems like the correct behaviour is for `allowWhitelisted` to clear any flags on the `IpRecord` whenever there's a match.

philbooth:
Fixes #205.

Whitelisted email addresses were setting the rate limit flag on the cached ip record. This meant that `/checkIpOnly` was being rate-limited for earlier actions performed by whitelisted email addresses, because it didn't have an email address of its own to whitelist against.

This PR fixes that by clearing `IpRecord::rl` after whitelisting in `/check`.  To get there it required a little bit of jiggling about of the surrounding code. I tried to do that in the least disruptive way, because this is the customs server and I have no idea what I'm doing. :smile:

@mozilla/fxa-devs r?

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -141,18 +141,25 @@ module.exports = function createServer(config, log) {
     // that we won't block. These are typically for Mozilla QA purposes.
     // They should be checked after everything else so as not to pay the
     // overhead of checking the many requests that are *not* QA-related.
-    if (result.block || result.suspect) {
-      if (ip in allowedIPs.ips || (email && allowedEmailDomains.isAllowed(email))) {
-        log.info({
-          op: 'request.check.allowed',
-          ip: ip,
-          block: result.block,
-          suspect: result.suspect
-        })
-        result.block = false
-        result.suspect = false
-      }
+    if ((result.block || result.suspect) && isWhitelisted(ip, email)) {
+      log.info({
+        op: 'request.check.allowed',
+        ip: ip,
+        block: result.block,
+        suspect: result.suspect
+      })
+
+      result.block = false
+      result.suspect = false
+
+      return true
     }
+
+    return false
+  }
+
+  function isWhitelisted (ip, email) {
+    return ip in allowedIPs.ips || (email && allowedEmailDomains.isAllowed(email))
   }
 
   function optionallyReportIp (result, ip, action) {
@@ -196,7 +203,7 @@ module.exports = function createServer(config, log) {
       fetchRecords(ip, email, phoneNumber)
         .spread(
           function (ipRecord, reputation, emailRecord, ipEmailRecord, smsRecord) {
-            if (ipRecord.isBlocked()) {
+            if (ipRecord.isBlocked() && ! isWhitelisted(ip, email)) {
               // a blocked ip should just be ignored completely
               // it's malicious, it shouldn't penalize emails or allow
               // (most) escape hatches. just abort!
@@ -253,24 +260,28 @@ module.exports = function createServer(config, log) {
               blockReason = blockReasons.IP_BAD_REPUTATION
             }
 
+            // Must be defined before the call to `allowWhitelisted`,
+            // because that function modifies `result` in-place.
+            const result = {
+              block: block,
+              blockReason: blockReason,
+              retryAfter: retryAfter,
+              unblock: canUnblock,
+              suspect: suspect
+            }
+
+            if (allowWhitelisted(result, ip, email)) {
+              // Prevent white-listed email addresses from blocking
+              // subsequent requests to /checkIpOnly
+              ipRecord.clear()
+            }
+
             return setRecords(ip, ipRecord, email, emailRecord, ipEmailRecord, phoneNumber, smsRecord)
-              .then(
-                function () {
-                  return {
-                    block: block,
-                    blockReason: blockReason,
-                    retryAfter: retryAfter,
-                    unblock: canUnblock,
-                    suspect: suspect
-                  }
-                }
-              )
+              .then(() => result)
           }
         )
         .then(
           function (result) {
-            allowWhitelisted(result, ip, email)
-
             log.info({
               op: 'request.check',
               email: email,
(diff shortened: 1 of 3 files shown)


lib/server.js
```diff
@@ -141,18 +141,27 @@ module.exports = function createServer(config, log) {
...
-        log.info({
-          op: 'request.check.allowed',
-          ip: ip,
-          block: result.block,
-          suspect: result.suspect
-        })
-        result.block = false
-        result.suspect = false
-      }
+    if ((result.block || result.suspect) && isWhitelisted(ip, email)) {
```
philbooth: The second test has been pulled out to `isWhitelisted` so that it can be re-used elsewhere.

lib/server.js
```diff
@@ -253,24 +262,26 @@ module.exports = function createServer(config, log) {
...
-                    suspect: suspect
-                  }
-                }
-              )
+              .then(() => result)
           }
         )
         .then(
           function (result) {
-            allowWhitelisted(result, ip, email)
```
philbooth: This had to move to before `setRecords` because we want to write the cleared ip record to the cache.

lib/server.js
```diff
@@ -196,7 +205,7 @@ module.exports = function createServer(config, log) {
       fetchRecords(ip, email, phoneNumber)
         .spread(
           function (ipRecord, reputation, emailRecord, ipEmailRecord, smsRecord) {
-            if (ipRecord.isBlocked()) {
+            if (ipRecord.isBlocked() && ! isWhitelisted(ip, email)) {
```
philbooth: The extra check is needed because we don't call `allowWhitelisted` in the following promise handler any more.

test/remote/check_tests.js
```diff
@@ -88,6 +98,49 @@ test(
   }
 )
 
+test('allowed email addresses in /check do not block subsequent requests to /checkIpOnly', t => {
```
philbooth: This test mimics what was happening in the functional tests when we noticed the problem.

lib/server.js
```diff
@@ -141,18 +141,29 @@ module.exports = function createServer(config, log) {
...
+      result.suspect = false
+
+      return true
+    }
+
+    return false
+  }
+
+  function isWhitelisted (ip, email) {
+    if (ip in allowedIPs.ips || (email && allowedEmailDomains.isAllowed(email))) {
```
seanmonstar: A `if bool { return true } else { return false }` can just be `return bool`. 😄

lib/server.js
seanmonstar: I personally always find the dangling bang odd, as if it's unloved. Grumble mumble spaces vs tabs code styles ignore me.

lib/server.js
```diff
@@ -253,24 +264,26 @@ module.exports = function createServer(config, log) {
               blockReason = blockReasons.IP_BAD_REPUTATION
             }
 
+            const result = {
```
seanmonstar: Any reason for moving this up?

I suppose a pedant could point out that before, a JSObject is only created on success, but I don't think it matters (this isn't a tight loop or anything, so shut up Mr Pedant).

philbooth: It's because `allowWhitelisted` has the hidden and surprising side-effect of modifying `result` in place. I'm definitely not a fan of side-effects so would be happy to change this to something less hidden and surprising. A clearer function name, returning the `result` maybe. Let me know if you have other suggestions.

seanmonstar: Oh I didn't realize. Perhaps just a comment to that effect?

lib/server.js
philbooth: You'll have to pry that dangling bang from my cold, dead fingers! There's plenty of love for it round my neck of the woods!

philbooth:
I should also add, it's possible this is the wrong way to fix the problem. I'm conscious that I lack understanding for lots of stuff in this repo, so I just kind of went with me gut. If clearing the `rl` flag is wrong or unsafe, I'm happy to do something else.

rfk:
Could this be used by a bad actor who controls an allowlisted email address, to hack around the IP-based ratelimiting in cases where we don't want them to?

lib/server.js
```diff
@@ -141,18 +141,25 @@ module.exports = function createServer(config, log) {
...
+      result.block = false
+      result.suspect = false
+
+      return true
     }
+
+    return false
+  }
+
+  function isWhitelisted (ip, email) {
```
rfk: FWIW, a while ago we invested a bit of effort into replacing "whitelisted" with other adjectives like "trusted" or "allowlisted" in service of inclusivity; let's maintain that maintain that state unless we have a good reason not to.

rfk:
> a bad actor who controls an allowlisted email address

This is potentially anyone, since IIRC @restmail.net is allowlisted in production.  The sort of thing I'm worried about here would be along these lines:

* Do some sort of bad action (e.g. password guessing) until my IP gets rate-limited
* Do a login attempt with restmail address, which clears the rate-limiting flag on my IP
* Resume doing the bad action now that the flag has been cleared
* Repeat indefinitely

It's not obvious whether this change would allow that or not.  I think it might avoid it for password-checking actions, but allow it for account-status-checking actions because `this.as` gets reset to `[]` inside `this.rateLimit()`.

If we want to go this route, we need the behaviour here to be obvious.

philbooth:
> @restmail.net is allowlisted in production.

Oh, I didn't realise that, sorry. I think we should close this down and do something else in that case.

PR #207, opened 2017-06-27
(iprecord): ignore prior rate-limiting for account access actions

### LINKED ISSUE #205: Current implementation of allowWhitelisted causes problems for /checkIpOnly

philbooth:
Fixes #205. Supercedes #206.

When running the content server functional tests, lots of requests to `/check` originate from the same IP address. These cause the rate limit flag to be set in `IpRecord`, but the request itself is not blocked because the email address matches the test in `allowWhitelisted`.

This is all fine until requests to `/checkIpOnly` start turning up. Now, the `IpRecord` is still marked as rate-limited when it is fetched from memcached, but we have no email address to whitelist on. So those requests get blocked, even though the specific action they're checking is below its threshold.

This PR fixes the issue using @rfk's preferred fix from https://github.com/mozilla/fxa-customs-server/issues/205#issuecomment-310551201, returning zero from `IpRecord::update` if the `aa` array is not over its threshold. Effectively account access actions are counted in a separate bucket to everything else.

@mozilla/fxa-devs r?

### DIFF
--- a/test/remote/check_tests.js
+++ b/test/remote/check_tests.js
@@ -1,13 +1,26 @@
 /* Any copyright is dedicated to the Public Domain.
  * http://creativecommons.org/publicdomain/zero/1.0/ */
 
-var test = require('tap').test
-var restify = require('restify')
-var TestServer = require('../test_server')
-var mcHelper = require('../memcache-helper')
+'use strict'
 
-var TEST_EMAIL = 'test@example.com'
-var TEST_IP = '192.0.2.1'
+const mcHelper = require('../memcache-helper')
+const Promise = require('bluebird')
+const restify = require('restify')
+const test = require('tap').test
+const TestServer = require('../test_server')
+
+const TEST_EMAIL = 'test@example.com'
+const TEST_IP = '192.0.2.1'
+const ALLOWED_EMAILS = []
+for (let i = 0; i < 3; ++i) {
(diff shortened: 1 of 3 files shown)


test/remote/check_tests.js
```diff
@@ -88,6 +102,81 @@ test(
...
+  })
+  .spread((req, res, obj) => {
+    t.equal(res.statusCode, 200, '/checkIpOnly succeeded')
+    t.equal(obj.block, false, 'request was not blocked')
+  })
+  .catch(err => t.fail(err))
+  .then(() => t.end())
+})
+
+test('disallowed email addresses in /check do block subsequent requests to /checkIpOnly', t => {
```
rfk: Should this read "do not block" rather than "do block"?  The test deems to assert that they're not blocked.

rfk:
This looks good to me with that one nit, r+

PR #208, opened 2017-06-28
feat(node): upgrade to Node 6

seanmonstar:


vladikoff:
Seems like it is working!

PR #216, opened 2017-11-13
fix(newrelic): newrelic native requires python, gyp, c++; update node@6.12.0

jrgm:
r? - @vladikoff


Dockerfile-build
```diff
@@ -1,4 +1,6 @@
-FROM node:6.11.5-alpine
+FROM node:6.12.0-alpine
+
+RUN apk add --no-cache make git gcc g++ python
```
vladikoff: 👍

PR #218, opened 2017-11-15
feat(sms): Add an allow list of non-rate limited phone numbers for testing.

### LINKED ISSUE #217: Allow list of test phone numbers that will not be rate limited.
Right now the content-server has a bunch of SMS related functional tests that aren't run on "prod-like" environments because they send actual SMSs. Instead of using randomly generated phone numbers, I tried using test phone numbers provided by phone companies, but was soon rate limited.

We should allow SMS messages to be sent to these special phone numbers w/o being rate limited, perhaps keeping a list of numbers or regexps that aren't rate limited - sort of like we do with email addresses.

Ref mozilla/fxa-content-server#5689
Replaces https://github.com/mozilla/fxa-auth-server/issues/2220

shane-tomlinson:
fixes #217

@rfk - r?

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -131,22 +131,10 @@ module.exports = function createServer(config, log) {
     return P.all(promises)
   }
 
-  function isIpAllowed(ip) {
-    return ip in allowedIPs.ips
-  }
-
-  function isEmailAllowed(email) {
-    return email && allowedEmailDomains.isAllowed(email)
-  }
-
-  function isPhoneNumberAllowed(phoneNumber) {
-    return phoneNumber && allowedPhoneNumbers.isAllowed(phoneNumber)
-  }
-
   function isAllowed(ip, email, phoneNumber) {
-    return isIpAllowed(ip) ||
-           isEmailAllowed(email) ||
-           isPhoneNumberAllowed(phoneNumber)
+    return allowedIPs.isAllowed(ip) ||
+           allowedEmailDomains.isAllowed(email) ||
+           allowedPhoneNumbers.isAllowed(phoneNumber)
   }
 
   function allowWhitelisted (result, ip, email, phoneNumber) {
--- a/lib/server.js
+++ b/lib/server.js
@@ -46,12 +46,14 @@ module.exports = function createServer(config, log) {
   var limits = require('./settings/limits')(config, Settings, log)
   var allowedIPs = require('./settings/allowed_ips')(config, Settings, log)
   var allowedEmailDomains = require('./settings/allowed_email_domains')(config, Settings, log)
+  const allowedPhoneNumbers = require('./settings/allowed_phone_numbers')(config, Settings, log)
   var requestChecks = require('./settings/requestChecks')(config, Settings, log)
 
   if (config.updatePollIntervalSeconds) {
     [
       allowedEmailDomains,
       allowedIPs,
+      allowedPhoneNumbers,
       limits,
       requestChecks
     ].forEach(settings => {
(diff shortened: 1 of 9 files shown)


lib/settings/allowed_phone_numbers.js
```diff
@@ -0,0 +1,41 @@
...
+      super('allowedPhoneNumbers')
+      this.setAll(phoneNumbers)
+    }
+
+    isAllowed(phoneNumber) {
+      return phoneNumber in this.phoneNumbers
+    }
+
+    setAll(phoneNumbers) {
+      this.phoneNumbers = {}
```
shane-tomlinson: I wasn't sure if a JavaScript Set is allowed in the customs-server, seems like it'd be a natural fit.

philbooth: I say do it, good shout!

lib/server.js
```diff
@@ -129,13 +131,32 @@ module.exports = function createServer(config, log) {
     return P.all(promises)
   }
 
-  function allowWhitelisted (result, ip, email) {
-    // Regardless of any preceding checks, there are some IPs and emails
-    // that we won't block. These are typically for Mozilla QA purposes.
-    // They should be checked after everything else so as not to pay the
-    // overhead of checking the many requests that are *not* QA-related.
+  function isIpAllowed(ip) {
```
vbudhram: With the introduction of these, wonder if we should we pull configs and logic into a common helper class and export `allowWhitelisted`?

lib/server.js
```diff
@@ -262,7 +283,7 @@ module.exports = function createServer(config, log) {
         )
         .then(
           function (result) {
-            allowWhitelisted(result, ip, email)
+            allowWhitelisted(result, ip, email, phoneNumber)
```
vbudhram: Can probably add a `getAllowSMS` endpoint similar to [these](https://github.com/mozilla/fxa-customs-server/blob/78226031d44e79064ac50a79c696553ff8540155/lib/server.js#L609).

lib/server.js
shane-tomlinson: Does `isAllowed` on `allowedEmailDomains` and `allowedPhoneNumbers` fit that?

lib/server.js
shane-tomlinson: On it.

shane-tomlinson:
> couldn't find anything on Nexmo's site, but can we send sms to certain numbers for free?

We switched to Amazon a few months back, I couldn't find anything on their docs either.

lib/server.js
```diff
@@ -129,13 +131,20 @@ module.exports = function createServer(config, log) {
...
+
+  function allowWhitelisted (result, ip, email, phoneNumber) {
+    // Regardless of any preceding checks, there are some IPs, emails,
+    // and phone numbers that we won't block. These are typically for
+    // Mozilla QA purposes. They should be checked after everything
+    // else so as not to pay the overhead of checking the many requests
+    // that are *not* QA-related.
     if (result.block || result.suspect) {
-      if (ip in allowedIPs.ips || (email && allowedEmailDomains.isAllowed(email))) {
+      if (isAllowed(ip, email, phoneNumber)) {
```
rfk: Nice cleanup 👍

scripts/config-list.js
```diff
@@ -41,6 +41,7 @@ var output = {
   limits: {},
   allowedIPs: [],
   allowedEmailDomains: [],
+  allowedPhoneNumbers: [],
```
rfk: IIUC, I think we also need to add logic to actually fetch `allowedPhoneNumbers` from memcache, similar to `mc.getAsync('allowedEmailDomains')`?

shane-tomlinson: [Added](https://github.com/mozilla/fxa-customs-server/pull/218/files#diff-51918078d313a138e042691cb119ed4dR79) - mind having a look?

PR #225, opened 2018-01-05
feat(token): Add rate limit for tokenCodes

### LINKED ISSUE #222: Support rate limiting verifyTokenCodes
From https://github.com/mozilla/fxa-auth-server/pull/2218/files, we need to add rate limits for verifying token codes in https://github.com/mozilla/fxa-customs-server/blob/08fca60761c9fdd987ef12dbb737410aec84c5ee/lib/actions.js#L18

New action is `verifyTokenCodes` and should use default rate limits when verifying codes.

vbudhram:
Fixes #222 

Opted to do a dynamic test to avoid duplication. Also took the chance to update to ES6 [syntax](https://github.com/mozilla/fxa-customs-server/pull/225/commits/02acdb2daef2d8f63998e251f4796976e9c773ee).

### DIFF
--- a/test/remote/too_many_verify_codes.js
+++ b/test/remote/too_many_verify_codes.js
@@ -1,15 +1,15 @@
 /* Any copyright is dedicated to the Public Domain.
  * http://creativecommons.org/publicdomain/zero/1.0/ */
 
-var test = require('tap').test
-var TestServer = require('../test_server')
-var Promise = require('bluebird')
-var restify = Promise.promisifyAll(require('restify'))
-var mcHelper = require('../memcache-helper')
+const test = require('tap').test
+const TestServer = require('../test_server')
+const Promise = require('bluebird')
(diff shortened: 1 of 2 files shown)


vbudhram:
@mozilla/fxa-devs r?

test/remote/too_many_verify_codes.js
```diff
@@ -22,54 +21,50 @@ process.env.RATE_LIMIT_INTERVAL_SECONDS = 1
...
-      t.type(testServer.server, 'object', 'test server was started')
+test('startup', (t) => {
+  testServer.start((err) => {
+    t.type(testServer.server, 'object', 'test server was started')
+    t.notOk(err, 'no errors were returned')
+    t.end()
+  })
+})
+
+const testCasesVerifyCode = [
```
rfk: style nit: I wonder if this would be clearer as `const VERIFY_CODE_ACTIONS` or similar

PR #227, opened 2018-02-06
Check ip only doc and log fix

g-k:
Changes:

* Add API docs for `POST /checkIpOnly`
* Fix an apparent typo in the op for the missing params error log

Checked that:

* the example from the docs `curl -w '\n' -v -H "Content-Type: application/json" "http://127.0.0.1:7000/checkIpOnly" -d '{"ip":"192.0.2.1","action":"accountCreate"}'` returns the expected result
* ran tests locally, a bunch fail but I don't think it's related to these changes

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -27,6 +27,7 @@ content-type of "application/json".
 * [POST /blockIp](#post-blockip)
 * [POST /check](#post-check)
 * [POST /checkAuthenticated](#post-checkauthenticated)
+* [POST /checkIpOnly](#post-checkiponly)
 * [POST /failedLoginAttempt](#post-failedloginattempt)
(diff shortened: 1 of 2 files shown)


vladikoff:
> ran tests locally, a bunch fail but I don't think it's related to these changes

Might need to start memcached

g-k:
With memcached running on 11211, local tests pass but some remote tests timed out so maybe I'm not running something else.

Tests passed in CI though.

philbooth:
Thanks @g-k!

vladikoff:
Filed https://github.com/mozilla/fxa-customs-server/issues/228 to hopefully fix bad test results in the future

Issue #228, opened 2018-02-07
Start memcached if not running as part of `npm test`


vladikoff:
Ref: https://github.com/mozilla/fxa-customs-server/pull/227#issuecomment-363468963

g-k:
Exiting and telling the dev to make sure memcached is running would work too.

vladikoff:
> Exiting and telling the dev to make sure memcached is running would work too.

Good backup solution :)

vladikoff:
cc @deeptibaghel 

This is very close to https://github.com/mozilla/fxa-local-dev/pull/100

If you run `npm test` without memcached right now a lot of tests will fail. 
Once we land a fix in https://github.com/mozilla/fxa-local-dev/pull/100 then it will be okay, but some devs do not use fxa-local-dev, so we need to add logic to start memcached into this project.

deeptibaghel:
@vladikoff I would like to work on this.

vladikoff:
@deeptibaghel sounds good, see this file https://github.com/mozilla/fxa-customs-server/blob/master/test/memcache-helper.js I Think if you make a fix in there, it should run the `spawn` command and run memcached server (if it is not running already) when that file is required.

vladikoff:
@deeptibaghel from the looks of it you might be better off using https://nodejs.org/api/child_process.html#child_process_child_process_spawnsync_command_args_options  (synchronous call to start the server)

and run that before this line https://github.com/mozilla/fxa-customs-server/blob/master/test/memcache-helper.js#L40

PR #230, opened 2018-03-14
fix(test): #228 Start memcached as part of 'npm test'

deeptibaghel:
Have added the lines to start memcached, not checking if it is running already as it does nothing if it is already running. spawnSync makes it waiting forever so i am using spawn only.

### DIFF
--- a/test/memcache-helper.js
+++ b/test/memcache-helper.js
@@ -10,8 +10,7 @@ P.promisifyAll(Memcached.prototype)
 
 var config = {
   memcache: {
-    host: '127.0.0.1',
-    port: '11211'
+    address: '127.0.0.1:11211',
   },
   limits: {
     blockIntervalSeconds: 1,
--- a/README.md
+++ b/README.md
@@ -14,13 +14,15 @@ Clone the git repository and install dependencies:
     cd fxa-customs-server
     npm install
 
-You'll also need to [install memcached](http://www.memcached.org/downloads),
-otherwise all requests will be blocked.
-By default, the customs server tries to connect to memcached
-using port `11211` on `127.0.0.1`.
-You can specify a different port and IP address
-using the `memcache.address` configuration setting
-or the `MEMCACHE_ADDRESS` environment variable.
+Install memcached
+
+    You'll need to [install memcached](http://www.memcached.org/downloads),
+    otherwise all requests will be blocked.
+    By default, the customs server tries to connect to memcached
+    using port `11211` on `127.0.0.1`.
+    You can specify a different port and IP address
+    using the `memcache.address` configuration setting
+    or the `MEMCACHE_ADDRESS` environment variable.
 
 To start the server, run:
 
@@ -39,12 +41,6 @@ To run the customs server via Docker:
 Run tests with:
 
     npm test
-
-On Mac OS X, memcached must be manually started for the tests to run.
-
-    memcached &
-    npm test
-
 
 To run tests via Docker:
 
--- a/scripts/test-local.sh
+++ b/scripts/test-local.sh
@@ -11,4 +11,13 @@ if test "$NO_COVERAGE" = ""; then
   cov="--coverage --cov"
 fi
 
+if ! echo stats | nc localhost 11211 | grep -q 'STAT'; then
+  memcached &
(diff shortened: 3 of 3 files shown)


deeptibaghel:
@vladikoff i am not clear on which test has timed out in travis

vladikoff:
hm in the end it shows: "test/local/ban_tests.js timeout!:
     timeout!
  "

test/memcache-helper.js
```diff
@@ -37,6 +36,9 @@ var config = {
   }
 }
 
+const spawn = require('child_process').spawn
+spawn('memcached', ['-v'])
```
vladikoff: try using spawnSync? it might now spawn fast enough for tests asynchronously

test/memcache-helper.js
```diff
@@ -10,8 +10,7 @@ P.promisifyAll(Memcached.prototype)
 
 var config = {
   memcache: {
-    host: '127.0.0.1',
-    port: '11211'
+    address: '127.0.0.1:11211',
```
vladikoff: what does this changed do?

deeptibaghel: 'address' is being used here but never defined :
https://github.com/mozilla/fxa-customs-server/blob/be3b5794da2df8d8d7d042a269ef3ca47e248c1f/test/memcache-helper.js#L41

vladikoff:
@deeptibaghel I also get this locally:

```
test/local/ban_tests.js ............................... 0/1 31s
  not ok timeout
```

checking now....

test/memcache-helper.js
```diff
@@ -37,6 +36,9 @@ var config = {
   }
 }
 
+const spawn = require('child_process').spawnSync
```
vladikoff: @deeptibaghel ah I see so we would need to use `spawn` here. I Think the tests hang because the process is left running after the tests are done in `ban_tests.js`

vladikoff: `spawnSync` just blocks the process and doesn't continue because the `memcached` just sits there

deeptibaghel:
@vladikoff but travis failure is not related to spawn i think. I added those 2 parameters but timeout still persists.

vladikoff:
@deeptibaghel Let's make a change in the README.md: 

https://github.com/mozilla/fxa-customs-server/blob/master/README.md#testing we can remove the 

> On Mac OS X, memcached must be manually started for the tests to run. 

section. 

Add a section into https://github.com/mozilla/fxa-customs-server/blob/master/README.md#development to mention that `memcached` needs to be installed

vladikoff:
We already have:

> You'll also need to install memcached, otherwise all requests will be blocked 

but we should move that around a bit, it is mentioned a bit late in the README.

vladikoff:
@deeptibaghel are you getting timeouts locally running:

```
 node_modules/.bin/tap -R spec test/local test/remote --coverage --cov
```

deeptibaghel:
> @deeptibaghel are you getting timeouts locally running:
> 
>  node_modules/.bin/tap -R spec test/local test/remote --coverage --cov

I am not getting any error for test/local/ban_tests.js.

deeptibaghel:
@vladikoff i have modified README.md

test/memcache-helper.js
```diff
@@ -37,6 +36,12 @@ var config = {
   }
 }
 
+const spawn = require('child_process').spawn
```
vladikoff: @deeptibaghel I talked to @vbudhram about this and we suspect `spawn` starts the process, but tests timeout because it is not cleaned up after the last test. We could write a function to stop the process, but we can just a simpler solution for now: Start memcached in this file https://github.com/mozilla/fxa-customs-server/blob/master/scripts/test-local.sh and run `kill` on it after tests are done. Edge case to test is the case where it is already running, see what happens.

Example: https://github.com/mozilla/fxa-auth-server/blob/master/scripts/start-local-mysql.sh#L8-L17

deeptibaghel:
@vladikoff ,  i think it's now working the way we wanted, the test script first checks if memcached is running, starts if not and kills later.

vladikoff:
@deeptibaghel Thank you!! 👍

PR #233, opened 2018-04-02
feat(totp): add totp rate limits

### LINKED ISSUE #226: TOTP Rate limiting codes
From security review we should rate limit these codes to 2 times every 30 seconds.

Ref: https://docs.google.com/document/d/1m4D_7DZAfXt4e-bEbfI-5s_x2XFO7TEQo1a_SQcj1O8/edit#

vbudhram:
Fixes #226 

This PR adds rate limiting to actions that involve checking a TOTP code (only `verifyTotpCode` atm ). Unfortunately, while adding this rate limit, I could not re-use the existing rate limiting structure for ip and email based actions since those limits are defined `config.limits.rateLimitIntervalSeconds`. The TOTP rate limit requirement is 2 checks every 30 seconds (rate the codes are generated).

To support this feature, I decided to build off https://github.com/mozilla/fxa-customs-server/pull/174, which had a lofty goal of refactoring all the things. This PR is a smaller subset that adds support for creating a custom rule for actions and checking limits.

* To maintain backwards comp, a new route is used `/checkWithRule`
* Added a generic record class that stores custom limits and actions for rule
* Records are stored in memcache with key `ip + email + ruleName`

While this doesn't negate the fact we should refactor this service, it does provide the minimum parts needed to support rate limiting TOTP.

@rfk What are your thoughts on this approach?

### DIFF
--- a/lib/config/config.js
+++ b/lib/config/config.js
@@ -279,6 +279,37 @@ module.exports = function (fs, path, url, convict) {
       default: '',
       format: 'String',
       env: 'SENTRY_DSN'
+    },
+    userDefinedRateLimitRules: {
+      totpCodeRules: {
+        actions: {
+          doc: 'Array of actions that this rule should be applied to',
+          default: [
+            'verifyTotpCode',
+          ],
+          format: Array
+        },
+        limits: {
+          max: {
+            doc: 'max actions during `period` that can occur before rate limit is applied',
+            format: 'nat',
+            default: 2,
+            env: 'TOTP_CODE_RULE_MAX'
+          },
+          periodMs: {
+            doc: 'period needed before rate limit is reset',
+            format: 'duration',
+            default: '30 seconds',
+            env: 'TOTP_CODE_RULE_PERIOD_MS'
+          },
+          rateLimitIntervalMs: {
+            doc: 'how long rate limit is applied',
+            format: 'duration',
+            default: '30 seconds',
+            env: 'TOTP_CODE_RULE_LIMIT_INTERVAL_MS'
+          }
+        }
+      }
     }
   })
 
--- a/lib/server.js
+++ b/lib/server.js
@@ -14,6 +14,7 @@ var blockReasons = require('./block_reasons')
 var P = require('bluebird')
 P.promisifyAll(Memcached.prototype)
 var Raven = require('raven')
+const utils = require('./utils')
 
 // Create and return a restify server instance
 // from the given config.
@@ -76,10 +77,29 @@ module.exports = function createServer(config, log) {
   var IpEmailRecord = require('./ip_email_record')(limits)
   var EmailRecord = require('./email_record')(limits)
   var IpRecord = require('./ip_record')(limits)
+  const Record = require('./record')
   var UidRecord = require('./uid_record')(limits)
   var smsRecord = require('./sms_record')(limits)
 
   var handleBan = P.promisify(require('./bans/handler')(config.memcache.recordLifetimeSeconds, mc, EmailRecord, IpRecord, log))
+
+  // Pre compute user defined rules into a Map<action, array of rule name>
+  function computeUserDefinedRules() {
+    const result = new Map()
+    const rules = config.userDefinedRateLimitRules
+    Object.keys(config.userDefinedRateLimitRules).forEach((key) => {
+      rules[key].actions.forEach((action) => {
+        const items = result.get(action)
+        if (! items) {
+          result.set(action, [key])
+        } else {
+          result.set(action, items.push(key))
+        }
+      })
+    })
+    return result
+  }
+  const computedRules = computeUserDefinedRules()
 
   var api = restify.createServer({
     formatters: {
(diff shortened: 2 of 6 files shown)


lib/config/config.js
```diff
@@ -279,6 +279,37 @@ module.exports = function (fs, path, url, convict) {
       default: '',
       format: 'String',
       env: 'SENTRY_DSN'
+    },
+    customRules: {
```
rfk: Naming nit: using "custom" in the sense of "customized" may be confusing when mixed with "customs" in the sense of "customs and border security", I wonder if we can find an equivalent term that's less ambiguous.

lib/config/config.js
```diff
@@ -279,6 +279,37 @@ module.exports = function (fs, path, url, convict) {
...
+            env: 'TOTP_CODE_RULE_MAX'
+          },
+          periodMs: {
+            doc: 'period needed before rate limit is reset',
+            format: 'duration',
+            default: '30 seconds',
+            env: 'TOTP_CODE_RULE_PERIOD_MS'
+          },
+          rateLimitIntervalMs: {
+            doc: 'how long long rate limit is applied',
```
rfk: nit: "long long"

lib/server.js
```diff
@@ -386,6 +387,91 @@ module.exports = function createServer(config, log) {
...
+      log.error({op: 'request.checkCustomRule', action, ip, email})
+      res.send(400, err)
+      return next()
+    }
+
+    // Get all the custom rules that might apply to this action
+    const customRules = config.customRules
+    const checkRules = Object.keys(customRules).filter((key) => {
+      return customRules[key].actions.indexOf(action) > -1
+    })
```
rfk: I suggest we pre-compute this at load time into a hash that maps action names to list of applicable rules, rather than doing a nested loop here on each request.

lib/server.js
```diff
@@ -386,6 +387,91 @@ module.exports = function createServer(config, log) {
...
+      const err = {code: 'MissingParameters', message: 'no rules found for ' + action}
+      log.error({op: 'request.checkCustomRule', action, ip, email})
+      res.send(400, err)
+      return next()
+    }
+
+    // Retrieve the record for each rule and check if `action` has been rate limited
+    const recordResults = []
+    checkRules.forEach((ruleName) => {
+      const recordKey = ip + email + ruleName
```
rfk: If it's a relatively small change, let's consider hashing this down to something with a known fixed length, to avoid https://bugzilla.mozilla.org/show_bug.cgi?id=1365486

lib/server.js
```diff
@@ -386,6 +387,91 @@ module.exports = function createServer(config, log) {
...
+    })
+    if (checkRules.length <= 0) {
+      const err = {code: 'MissingParameters', message: 'no rules found for ' + action}
+      log.error({op: 'request.checkCustomRule', action, ip, email})
+      res.send(400, err)
+      return next()
+    }
+
+    // Retrieve the record for each rule and check if `action` has been rate limited
+    const recordResults = []
```
rfk: The `P.each` helper might make this iteration cleaner.

vbudhram:
Pulling WIP off of this since I think it is ready. It can be tested against https://github.com/mozilla/fxa-auth-server/pull/2386, just make sure you comment out `customsUrl` in `dev.json` before running.

Currently, this rule will only apply when verifying a TOTP code and allows 2 attempts every 30s.

@mozilla/fxa-devs r?

lib/server.js
```diff
@@ -386,6 +406,88 @@ module.exports = function createServer(config, log) {
...
+    return P.each(checkRules, (ruleName) => {
+      let retryAfter = null
+      const recordKey = ip + utils.createHashHex(email, ruleName)
```
vbudhram: Is there any value in keeping IP as part of memcache key? For example, would we ever need to delete user defined rules for a specific IP?

lib/server.js
```diff
@@ -386,6 +406,88 @@ module.exports = function createServer(config, log) {
     }
   )
 
+  api.post('/checkCustomRule', function(req, res, next) {
```
rfk: nit: this api still says "custom rule", which no longer matches the terminology used in the code here.

lib/server.js
rfk: I can't think of a reason why we'd want to do this off the top of my head.  I also don't recall there being a way to delete keys from memcached by prefix, so in order to delete the key, you'd have to know all the inputs to the key calculation anyway.

I think we're more likely to want to have the ruleName in cleartext than either the IP or email, which would make this something like `recordKey = ruleName + ":" + utils.createHashHex(ip, email)` or similar.

lib/server.js
```diff
@@ -386,6 +406,88 @@ module.exports = function createServer(config, log) {
...
+          // actions since they can be referenced from config
+          record.limits = null
+          record.actions = null
+
+          const lifetime = Math.max(config.memcache.recordLifetimeSeconds, minLifetimeMS / 1000)
+          return mc.setAsync(recordKey, record, lifetime)
+        })
+        .then(() => {
+          recordResults.push({
+            block: retryAfter > 0,
```
rfk: Do you actually use this `block` key in the logic below?  I wonder if this would be simpler as just a list of `retryAfter` integers, rather than a list of rich objects with a `retryAfter` key.

lib/utils.js
```diff
@@ -0,0 +1,18 @@
...
+'use strict'
+
+const crypto = require('crypto')
+
+module.exports = {
+  createHashHex () {
+    const hash = crypto.createHash('sha256')
+    const args = [...arguments]
+    args.forEach((arg) => {
+      hash.update(arg)
```
rfk: Please include a separator between each argument here, preferably one that won't appear in any of the input strings, such as `'\00'`.  Otherwise, the hash for e.g. `{ ip: "1.2.3.45", email: "test@example.com }` could wind up being the same as the hash for `{ ip: "1.2.3.4", email: "5test@example.com" }`, which would cause us to incorrectly treat those as the same.

test/remote/too_many_totp_codes.js
```diff
@@ -0,0 +1,96 @@
...
+    .spread((req, res, obj) => {
+      t.equal(res.statusCode, 200, 'returns a 200')
+      t.equal(obj.block, false, 'not rate limited')
+      t.end()
+    })
+    .catch((err) => {
+      t.fail(err)
+      t.end()
+    })
+})
```
rfk: I think it'd be worth adding another test here, to ensure that the rate-limiting is independent of other rate-limiting actions.  For example, that a user can still attempt `accountLogin` actions after be rate-limited for totp verification, and conversely that a user rate-limited for `accountLogin` actions can still do totp verification attempts.

lib/server.js
rfk: By moving this into its own route, we bypass all the checks in the usual `/check` route.  IIUC that's mostly what we want, but I wonder if we've lost any that we should keep, such a the ip-reputation-service checking, the ip-blocklist checking, or the explicit "block this IP address" API call offered by the `/blockIp` route?

vbudhram:
@rfk After thinking it over, it seems reasonable to put this logic into the `/check` route. I updated the code to do that and refactored the route a little bit.

Mind another review at your convenience? Probably without [whitespace](https://github.com/mozilla/fxa-customs-server/pull/233/files?w=1).

lib/utils.js
```diff
@@ -0,0 +1,18 @@
...
+'use strict'
+
+const crypto = require('crypto')
+
+module.exports = {
+  createHashHex () {
+    const hash = crypto.createHash('sha256')
+    const args = [...arguments]
+    args.forEach((arg) => {
+      hash.update(arg + '\\00')
```
rfk: I think this should be `'\0'` in order to get a literal null byte, rather than the string "\00".

PR #234, opened 2018-04-11
fix(tests): fix dependent tests

vbudhram:
I was setting up a new computer yesterday and found that these tests depended on a previous tests state which caused some test failures. I suspect that the older computer was slow enough that the limits were reset before the next test. This fixes the issue my using unique ips, emails and actions when needed.

Unfortunately, there are probably more tests that have shared states but it would be more work to find and fix them than to do a complete refactor such as https://github.com/mozilla/fxa-customs-server/pull/174.

### DIFF
--- a/test/utils.js
+++ b/test/utils.js
@@ -1,0 +1,32 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
+const crypto = require('crypto')
+
+module.exports = {
+  getSubnet: function () {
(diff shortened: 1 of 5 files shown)


vbudhram:
@mozilla/fxa-devs r?

test/utils.js
```diff
@@ -0,0 +1,32 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
+const crypto = require('crypto')
+
+module.exports = {
+  getSubnet: function () {
+    return Math.floor(Math.random() * 255)
+  },
```
rfk: This seems to duplicate the `getSubnet` inner function defined in `randomIp` below?

PR #237, opened 2018-04-18
chore(docs): Add documentation on tagging private releases.

rfk:
I do a bunch of private-repo wrangling when tagging customs-server that's probably not obvious to others, so I thought I better write it down.  @mozilla/fxa-devs r?

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,48 @@ To run tests via Docker:
 docker-compose run mozilla/fxa_customs_server npm test
 ```
 
+## Tagging Releases
+
+Unlike other FxA services,
+the customs-server includes some non-public code changes
+that are managed in a separate private repo:
+
+  https://github.com/mozilla/fxa-customs-server-private/
+
+When tagging a new release for deployment,
+it needs to be merged to the private repo
+and re-tagged with those changes in place.
+The process looks something like this:
+
+```
+> # First, make a new public tag.
+> cd ./fxa-customs-server
+> grunt version
+> git push; git push --tags
+>
+> # Next, merge the updates to the private repo.
+> cd ../fxa-customs-server-private
+> git checkout master ; git pull
+> git remote add public https://github.com/mozilla/fxa-customs-server
+> git fetch public
+> git merge public/master
+> git push
+>
+> # Make a release branch in the private repo.
+> git checkout v1.XXX.0
+> git checkout -b train-XXX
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -48,6 +48,48 @@ To run tests via Docker:
...
+> # Make a release branch in the private repo.
+> git checkout v1.XXX.0
+> git checkout -b train-XXX
+> git push -u origin train-XXX
+>
+> # Merge private changes from previous train.
+> git merge origin/train-(XXX-1)
+>
+> # Make a private tag to include those changes.
+> git tag v1.XXX.0-private
```
vladikoff: nice! this is important to remember 👍

rfk:
Naturally, it would be better if we could move the private bits into config rather than code so that we didn't have to make private releases at all, but at the very least it shouldn't be all in my head...

README.md
```diff
@@ -48,6 +48,48 @@ To run tests via Docker:
...
+> git push; git push --tags
```
vladikoff: I wonder if some super rare git configurations or a bad origin setup this might push a private tag to a public repo by accident, but i might be over cautious

rfk: Yeah, I recommend always checking out the private repo separately so that it's `origin`, I've had a few accidents when trying to manage it as a secondary remote in an otherwise public repo :-(

philbooth: I've had those accidents before too and never use `git push --tags` any more, I always push the specific tag I want explicitly. But I am quite clumsy, so maybe that level of precaution is not for everyone.

README.md
```diff
@@ -48,6 +48,48 @@ To run tests via Docker:
...
+Unlike other FxA services,
+the customs-server includes some non-public code changes
+that are managed in a separate private repo:
+
+  https://github.com/mozilla/fxa-customs-server-private/
+
+When tagging a new release for deployment,
+it needs to be merged to the private repo
+and re-tagged with those changes in place.
+The process looks something like this:
```
philbooth: Fwiw, I think this would be more useful it was a shell script than a command listing in a readme.

README.md
rfk: This sounds like a reasonable precaution FWIW

shane-tomlinson: I have a script that I use to do this:

> git push origin refs/tags/$TAG_NAME --no-verify

That avoid pushing unexpected tags.

PR #249, opened 2018-06-07
fix(deps): update ip rep client to v4

g-k:
Update ip-reputation-js-client to v4 which includes the changes:

* remove the add method (I don't see customs-server using this anywhere)
* downcased field names (fixed Reputation to reputation and Violation to violation; don't see any other occurrences)
* switch to the memcached backed [iprepd](https://github.com/mozilla-services/iprepd) backend
* add eslint changes and fixes

cc @ameihm0912 who wants to make sure the client tracks failed requests and latency (in this PR or a new separate one)

### DIFF
--- a/lib/reputationService.js
+++ b/lib/reputationService.js
@@ -30,8 +30,8 @@ var get = function (log, ipClient, ip) {
   return ipClient.get(ip)
     .then(function (response) {
       if (response && response.body && response.statusCode === 200) {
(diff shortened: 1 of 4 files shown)


g-k:
This will also require ops to switch the reputation service URL to `iprepd.{stage.prod}.mozaws.net` (the hawk creds are the same).

tigerblood.stage.mozaws.net is currently a CNAME to iprepd but prod is running them side by side.

PR #252, opened 2018-06-12
fix(tests): fix test-local.sh bug introduced in d177360

ameihm0912:
See also #250

### DIFF
--- a/scripts/test-local.sh
+++ b/scripts/test-local.sh
@@ -18,6 +18,6 @@ fi
 
 tap test/local test/remote $cov
 
-if [ -n $MC ]; then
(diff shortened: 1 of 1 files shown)


ameihm0912:
@vladikoff I noticed that change in #250 looks to have caused some issues with the Circle build, this should fix that

jrgm:
@shane-tomlinson - can you merge this into master. I can't build an update to node version without it.

shane-tomlinson:
@jrgm - if you can r+ https://github.com/mozilla/fxa-customs-server/pull/254, this will be merged into master.

PR #254, opened 2018-06-13
merge train-114 into master

shane-tomlinson:
@mozilla/fxa-devs - r?


rfk:
Circle is failing, this may need the fix from https://github.com/mozilla/fxa-customs-server/pull/252 merged and backported to the tag.

shane-tomlinson:
@jrgm - r?

PR #255, opened 2018-06-13
fix(docker): base image node:8-alpine and upgrade to npm6

jrgm:
r? - @vladikoff


jrgm:
circle-ci fails without https://github.com/mozilla/fxa-customs-server/pull/252

PR #263, opened 2018-07-31
fix(deps): update ip-reputation-js-client to 4.1.0

g-k:
Makes timeout param include DNS resolution and `connect()` not just time since last byte after a connection is established.

https://github.com/mozilla-services/ip-reputation-js-client/pull/24


g-k:
Thanks! We might need to bump up the timeout.

vladikoff:
ok!

PR #277, opened 2018-09-06
feat(code): add custom rate limits for token code

vbudhram:
Connects to https://github.com/mozilla/fxa-content-server/issues/5990

This PR adds the custom rate limit rule for `verifyTokenCode`. The limit is 5 attempts every 15mins, after which you are blocked for 15mins.


test/remote/custom_rules.js
```diff
@@ -0,0 +1,135 @@
+/* Any copyright is dedicated to the Public Domain.
+ * http://creativecommons.org/publicdomain/zero/1.0/ */
```
rfk: A tiny nit: I think it would be clearer if this file were named "user_defined_rules.js" to match the name of the config item it's testing.  "Custom" is too close to "customs" and could cause confusion.

Issue #271, opened 2018-08-22
Add `npm audit` support


shane-tomlinson:
Run in travis, do not fail at first.

ref #270

Issue #270, opened 2018-08-22
Remove nsp


shane-tomlinson:
nsp is being deprecated on Sept 30.

ref #271

PR #280, opened 2018-09-26
fix(deps): add filtered npm audit support

g-k:
refs: #270, #271

https://github.com/mozilla/fxa/issues/303
https://bugzilla.mozilla.org/show_bug.cgi?id=1478025

edit: this could be a grunt task too though it's been awhile since I've written grunt tasks.


package.json
```diff
@@ -36,6 +37,7 @@
     "restify-safe-json-formatter": "0.3.0"
   },
   "devDependencies": {
+    "audit-filter": "^0.3.0",
```
shane-tomlinson: Can you remove the "^" prefix here?

PR #317, opened 2019-01-31
feat(fraud-detection): Listen for DataFlow fraud detection events

### LINKED ISSUE #311: Add GCP Pub/Sub consumer to listen for fraud detection events
This is to consume fraud detection events generated by the pipeline @ameihm0912 and @moz-jvehent are creating. 

The goal of this bug is only to consume events without processing them to ensure we can test end to end data flow.

shane-tomlinson:
Hook up to GCP PubSub to listen for fraud detection events
from the DataFlow fraud detection pipeline.

fixes #311

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -16,11 +16,13 @@ var P = require('bluebird')
 P.promisifyAll(Memcached.prototype)
 var Raven = require('raven')
 const utils = require('./utils')
+const dataflow = require('./dataflow')
 
 // Create and return a restify server instance
 // from the given config.
 
 module.exports = function createServer(config, log) {
+  dataflow(config, log)
 
   var startupDefers = []
 
--- a/test/local/dataflow_tests.js
+++ b/test/local/dataflow_tests.js
@@ -1,0 +1,179 @@
+/* Any copyright is dedicated to the Public Domain.
+ * http://creativecommons.org/publicdomain/zero/1.0/ */
+
+const proxyquire = require('proxyquire')
+const sinon = require('sinon')
+const { test } = require('tap')
+
+let log
+let config
+let dataflow
+let PubSub
+let pubsub
+let subscription
+
+test(
+  'shared setup', async (t) => {
+    log =  {
+      error: sinon.spy(),
+      info: sinon.spy()
+    }
+
+    subscription = {
+      on: sinon.spy()
+    }
+
+    pubsub = {
+      subscription: sinon.spy(() => subscription)
+    }
+
+    PubSub = {
+      PubSub: sinon.spy(function (config) {
+        return pubsub
+      })
+    }
+
+    dataflow = proxyquire('../../lib/dataflow', {
+      '@google-cloud/pubsub': PubSub
+    })
+  }
+)
+
+test(
+  'setup not enabled causes a no-op', async (t) => {
+    config = {
+      dataflow: {
+        enabled: false,
+        gcpPubSub: {
+          projectId: 'foo',
+          subscriptionName: 'bar'
+        }
+      }
+    }
+
+    dataflow(config, log)
+
+    t.equal(PubSub.PubSub.called, false)
+  }
+)
+
+test(
+  'setup missing projectId', async (t) => {
+    config = {
+      dataflow: {
+        enabled: true,
+        gcpPubSub: {
+          subscriptionName: 'bar'
+        }
+      }
+    }
+
+    t.throws(() => dataflow(config, log))
+  }
+)
+
+test(
+  'setup missing subscriptionName', async (t) => {
+    config = {
+      dataflow: {
+        enabled: true,
+        gcpPubSub: {
+          projectId: 'foo'
+        }
+      }
+    }
+
+    t.throws(() => dataflow(config, log))
+  }
+)
+
+test(
+  'setup valid config', async (t) => {
+    config = {
+      dataflow: {
+        enabled: true,
+        gcpPubSub: {
+          projectId: 'foo',
+          subscriptionName: 'bar'
+        }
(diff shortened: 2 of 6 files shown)


shane-tomlinson:
Pulling WIP off of this. I set up a new project in the GCP console (stomlinson-fxa-dataflow-dev) and a new topic (dataflow), and a new subscription (fxa). I sent a message, and it came through:

screenshot 2019-02-27 at 12 11 30

Some questions:

1. Can we expect a project to be available? If not, not sure how to handle this.
2. Can we expect a topic/subscription to be available? If not, should we create one?
3. If we *can* expect a project/topic/subscription, should we fail hard if they aren't?

@mozilla/fxa-devs - r?

shane-tomlinson:
cc @ameihm0912

lib/server.js
```diff
@@ -16,12 +16,17 @@ var P = require('bluebird')
...
 var Raven = require('raven')
 const utils = require('./utils')
+const dataflow = require('./dataflow')
 
 // Create and return a restify server instance
 // from the given config.
 
 module.exports = function createServer(config, log) {
 
+  if (config.dataflow.enabled) {
```
philbooth: I notice we test this flag twice. You could just make this condiation a nop internally in `lib/dataflow`, then this code wouldn't have to know anything about the internal config of that module.

test/local/dataflow_tests.js
```diff
@@ -0,0 +1,162 @@
...
+
+let log
+let config
+let dataflow
+let PubSub
+let pubsub
+let subscription
+
+test(
+  'shared setup', async (t) => {
```
philbooth: How come `async`? I don't see any `await` or promise-related stuff here or in the module under test...

shane-tomlinson: Because of tap. If the function returns a promise, then tap does not require you to call t.end. However if you do not return a promise, you have to call t.end. It seemed cleaner this way.

lib/server.js
shane-tomlinson: I like this suggestion.

test/local/dataflow_tests.js
philbooth: Ah, makes sense, thanks.

lib/server.js
shane-tomlinson: Made that update.

shane-tomlinson:
> * Can we expect a project to be available? If not, not sure how to handle this.

I have added logging, if this happens an error is logged but it doesn't prevent the server from starting.

> * Can we expect a topic/subscription to be available? If not, should we create one?

Like above, I have added logging, if this happens an error is logged but it doesn't prevent the server from starting.

> * If we _can_ expect a project/topic/subscription, should we fail hard if they aren't?

I took the approach for testing, no.

> * Can subscriptions/projects ever go away unexpectedly? I'm wondering how the customs server should react if they can.

I tried this, by both deleting subscriptions and topics while the content server is running. No error is displayed by the customs server, I haven't dug in far to determine whether the GCP library surfaces this condition.

ameihm0912:
@shane-tomlinson 

> 1. Can we expect a project to be available? If not, not sure how to handle this.

This will already exist yeah, we have a project setup for production and nonprod use right now for this

> 2. Can we expect a topic/subscription to be available? If not, should we create one?

A topic will already exist. For the subscription, believe we have a couple options. If each instance customs server creates a subscription, they will all get a copy of every message published to the topic. If a subscription is created, and all customs server instances use that single subscription, they will load balance across (each customs instance will get a subset of the published messages).

> 3. If we _can_ expect a project/topic/subscription, should we fail hard if they aren't?

I am guessing here we would want to just log this as a warning and continue, but this scenario should result in an immediate escalation.

> 4. Can subscriptions/projects ever go away unexpectedly?

Projects will not unless they are deleted; likewise with the topics. The subscriptions by default expire after 31 days of inactivity (e.g., no active connections to one). The subscriptions I believe can be configured to persist however.

> 5. What's going to happen as soon as we enable this in prod, are we going to get a flood of events?

Messages sent to a topic with no subscriptions are just dropped. So in the case of enabling it, if it's a new subscription it would just start with any events published after that subscription was created.

> 6. Do we have to worry about rate limiting?

Pubsub will rate limit messages based on how quickly the subscriber is ack'ing them, so there shouldn't need to be any rate limiting in customs required. If it cannot process messages fast enough however it will result in a backlog, but given the expected message rate I don't see this being an issue.

@jbuck feel free to correct me if I'm off anywhere here

jbuck:
> > 1. Can we expect a topic/subscription to be available? If not, should we create one?
> 
> A topic will already exist. For the subscription, believe we have a couple options. If each instance customs server creates a subscription, they will all get a copy of every message published to the topic. If a subscription is created, and all customs server instances use that single subscription, they will load balance across (each customs instance will get a subset of the published messages).

I think it makes sense to assume that a subscription exists, and we'll configure the customs server to point at it so that messages are load balanced across all instances of the customs server

> > 2. If we _can_ expect a project/topic/subscription, should we fail hard if they aren't?
> 
> I am guessing here we would want to just log this as a warning and continue, but this scenario should result in an immediate escalation.

Agreed. Monitoring on subscription queue length/age should handle the case where the customs server is not consuming messages

> > 4. Can subscriptions/projects ever go away unexpectedly?
> 
> Projects will not unless they are deleted; likewise with the topics. The subscriptions by default expire after 31 days of inactivity (e.g., no active connections to one). The subscriptions I believe can be configured to persist however.

You could have API failures, but I think it'd be unlikely to actually be deleted in regular use

> @jbuck feel free to correct me if I'm off anywhere here

I think we're all in agreement here

PR #324, opened 2019-03-20
refactor(server): Extract and simplify record retreival/storage, user defined rules.

shane-tomlinson:
server.js has a whole bunch of mixed concerns, part of which was record retrieval
and loading/checking user defined rules.

This PR extracts record handling logic as well as user defined rules logic
into their own modules.

Loading/saving records can now be done through a common interface. fetchRecords no longer
holds the assumption that an ip address will be passed in. setRecord no longer requires
passing in a key as the key is stored on the record, and setRecords now only accepts records
instead of it's confusing signature.  It's now possible to define non-enumerable
properties on a record that are not saved when serialized.

I started to use async/await to simplify logic where it made sense as well as
started down the path to using native promises in places.

Note, no remote tests are modified, so functionality should be the same.

This is groundwork to simplify the DataFlow integration where a simple API is
needed to fetch records of varying types.

@mozilla/fxa-devs - r?

blocks #325

### DIFF
--- a/lib/bans/handler.js
+++ b/lib/bans/handler.js
@@ -2,74 +2,30 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-module.exports = function (LIFETIME_SEC, mc, EmailRecord, IpRecord, log) {
-
-  function blockIp(ip, cb) {
-    mc.get(ip,
-      function (err, data) {
-        if (err) {
-          log.error({ op: 'handleBan.blockIp', ip: ip, err: err })
-          return cb(err)
-        }
-
-        log.info({ op: 'handleBan.blockIp', ip: ip })
-        var ir = IpRecord.parse(data)
-        ir.block()
-        var lifetime = Math.max(LIFETIME_SEC, ir.getMinLifetimeMS() / 1000)
-        mc.set(ip, ir, lifetime,
-          function (err) {
-            if (err) {
-              log.error({ op: 'memcachedError', err: err })
-              return cb(err)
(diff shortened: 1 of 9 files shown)


lib/bans/handler.js
```diff
@@ -2,73 +2,29 @@
...
-      blockIp(message.ban.ip, cb)
-    }
-    else if (message.ban && message.ban.email) {
-      blockEmail(message.ban.email, cb)
-    }
-    else {
+      return blockIp(message.ban.ip)
+    } else if (message.ban && message.ban.email) {
+      return blockEmail(message.ban.email)
+    } else {
```
vbudhram: You could probably drop the else statement here and just return.

lib/records.js
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  /**
+   * Save records
+   *
+   * @param {Record[]} records to save
+   * @returns {Promise} Resolves when complete
+   */
+  function setRecords(...records) {
+    const promises = records.map(record => record && setRecord(record))
```
vbudhram: The `record && setRecord(record)` part seem strange to me, when would a record be undefined and have called `setRecords`?

lib/user_defined_rules.js
```diff
@@ -0,0 +1,63 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports = function (config, fetchRecord, setRecord) {
```
vbudhram: Thanks for pulling this out!

PR #326, opened 2019-03-20
feat(docker): Use node 10 to build the docker image

shane-tomlinson:


shane-tomlinson:
@mozilla/fxa-devs - r?

PR #331, opened 2019-04-02
fix(deps): Fix the audit warnings.

shane-tomlinson:
Update:
* tap
* js-yaml
* grunt


shane-tomlinson:
@mozilla/fxa-devs - r?

blocks #330

